PR #27, opened 2016-12-12
Travis

serin-delaunay:
Changes:
* Add Travis CI config.
* Add rapidjson submodule.
* CMake: Require SFML 2.2, C++14.
* Use `boost::filesystem::path::operator/=` instead of `::append` (which is not available by default on Trusty).

Status:
* Wangscape builds, WangscapeTest compiles but fails to link gtest.
* SFML is installed from binaries downloaded from http://sfml-dev.org.
* Rapidjson is installed from git.
* Libnoise is installed from apt.
* Gtest is installed from apt then compiled.

> Build currently fails due to Travis-CI not having rapidjson (https://github.com/travis-ci/apt-package-whitelist/issues/3878) and libnoise (https://github.com/travis-ci/apt-package-whitelist/issues/3877) whitelisted.
> 
> If it takes a long time for them to be approved, or they are rejected, it may be preferable to add them as submodules (libnoise as a separate assembly with dynamic linkage).


hryniuk:
Hey, these packages have opened issues with whitelisting for four days now 😆. As we agreed on **spotify-json** we could remove **rapidjson-dev** from `.travis.yml` and add libnoise as submodule (if I didn't miss anything, Wangscape builds without it, so for now we can remove it from `.travis.yml` as well).

OK?

Your project has already a community and it is more and more important to have automatic builds :wink:.

serin-delaunay:
Yep, we've waited long enough :D

serin-delaunay:
I've added rapidjson as a submodule. Now the situation is:
* Building on Precise fails because the CMake version is too low (and the current CMakeLists.txt does need post-2.8.7 features).
* Building on Trusty fails because SFML is not whitelisted.

The preferable solution would be to switch to Trusty, add SFML as a submodule, and tell Travis to build it.

hryniuk:
It's a pity that such big project as SFML will be Wangscape's submodule, but I see that's the only way to go 😞 for now (btw I'm still working on switching to spotify-json, should finish it on Wednesday).

serin-delaunay:
Yes, it's not ideal. Although with the [proper configuration](https://docs.travis-ci.com/user/caching/#Caching-directories-(Bundler%2C-dependencies)) it should be possible to cache SFML (and libnoise) builds on Travis, so the impact on normal build time should be minimised.

Now that Perlin tile generation has basic functionality I'm working on adding the submodules. Github Desktop seems to get confused by them :/

hryniuk:
Fine. I need still a while to finish switching to spotify-json, it should be done on Wednesday.

serin-delaunay:
Building SFML requires libudev-dev, which is unapproved on trusty since February (https://github.com/travis-ci/apt-package-whitelist/issues/2449). So, this will need to be built in the sudo-enabled environment, which will take longer to boot. It looks like the only submodule dependency then required would be spotify-json.

serin-delaunay:
OK, as of https://github.com/serin-delaunay/Wangscape/commit/d9af272c8c042f3431a2d0cafee780e90494ff74 I've added all the necessary submodules and APT packages. GTest needs special treatment - only source code is delivered, binaries have to be built on the host machine. Fortunately that's a quick process.

The CMake scripts seem to be incorrect, because the rapidjson folder is being detected as a compilation target and the rapidjson test suite is built.

Wangscape still won't build because `#include <rapidjson/document.h>` fails. The CMake scripts must be including the wrong directories.

I don't know how to fix either problem, but it's probably not worth fixing them in the rapidjson version, if spotify-json integration will be complete so soon.

serin-delaunay:
I fixed the missing rapidjson dependency by doing `make install`. This still wastes time building the rapidjson tests, but at least the headers are found.

However, the trusty SFML package is 2.1, which lacks sf::BlendMode, which is required for Wangscape. So, a manual install is needed. If we only do 64-bit builds, we can still use pre-built binaries from the SFML website. I'll try to use the same manual copy to install rapidjson without building tests.

serin-delaunay:
SFML and rapidjson are both detected now. Libnoise was whitelisted on Trusty, so it might be possible to migrate back to the container platform, if CMake can find SFML 2.4.1+ and the JSON library when they're installed in userspace.

Wangscape builds without any errors, but WangscapeTest fails when trying to link gtest.

hryniuk:
I will need a few more days to integrate `spotify-json`. One reason is that I have to find good way to parametrize JSONs with e.g. filenames, as they're used as keys in JSON objects like `TilesetData` in `MetaOutput`.

serin-delaunay:
That's fine :) My current task doesn't depend on the JSON functionality, and we're both volunteers, so there's no rush.

Is there anything that can be changed in the JSON I/O formats to make the task easier? At this stage of development preserving compatibility is not a priority.

hryniuk:
Filename could be a value instead of a key, but I have to check if it really makes parsing easier.
I mean changing:
```
"g.s.png": {
"terrains": ...
}
```
to something like:
```
"tileset-files": [{
"filename": "g.s.png",
"terrains": {
"parts": ["g", "s"],
...
}]
```

serin-delaunay:
That would be fine, maybe even an improvement.

hryniuk:
I'll try to make Travis work with the current _master_ branch.

hryniuk:
I've tried, but setting it on _precise_ or _trusty_ is too tedious for me :wink:. [Docker builds](https://docs.travis-ci.com/user/docker/) look promising and I'll try to do it with them.

PR #30, opened 2016-12-13
Refactor tileset generator

hryniuk:


Wangscape/TilesetGenerator.cpp
```diff
@@ -28,34 +28,21 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
...
     {
-        size_t res_x = options.resolution;
-        size_t res_y = options.resolution;
-        bool x_dimension = true;
-        for (int i = 0; i < CORNERS; i++)
-        {
-            size_t& res_z = x_dimension ? res_x : res_y;
-            res_z *= clique.size();
-            x_dimension = !x_dimension;
-        }
```
hryniuk: Here you just did:
```cpp
size_t& res_z = res_x;
res_z *= clique.size();
size_t& res_z = res_y;
res_z *= clique.size();
size_t& res_z = res_x;
res_z *= clique.size();
size_t& res_z = res_y;
res_z *= clique.size();
```

Wangscape/TilesetGenerator.cpp
```diff
@@ -28,34 +28,21 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
...
-        for (auto terrain : clique)
-        {
-            ss << terrain << ".";
-        }
-        ss << options.fileType;
-        std::string filename = ss.str();
+        size_t res_x = options.resolution * clique.size() * clique.size();
+        size_t res_y = options.resolution * clique.size() * clique.size();
+
+        std::unique_ptr<sf::RenderTexture> output{getBlankImage(res_x, res_y)};
```
hryniuk: Now this function has in its name the note from the comment above :smile:.

I think `getBlankImage` could receive some `Resolution` object with proper methods for width and height and I suppose SFML has already something like that, **but** we need to check if it really fits here.

Wangscape/TilesetGenerator.cpp
```diff
@@ -102,4 +89,23 @@ void TilesetGenerator::generateClique(const Options::Clique& clique, sf::RenderT
...
+std::string TilesetGenerator::getOutputImageFilename(const Options::Clique& clique) const
+{
+    std::stringstream ss;
+    for (auto terrain : clique)
+    {
+        ss << terrain << ".";
+    }
+    ss << options.fileType;
+    return ss.str();
+}
```
hryniuk: Oh, now it can be tested! I didn't think about it, need to add some basic tests.

Wangscape/TilesetGenerator.cpp
serin-delaunay: I agree that code needed some simplification. However, I would prefer to keep it dependent on `CORNERS` to reduce refactoring work in the future. https://github.com/hryniuk/Wangscape/pull/1/commits/bc790b57dcc4cdde19d62069da2be5141c79c830 and https://github.com/hryniuk/Wangscape/pull/1/commits/d8187380c2f896ae72ce0517d432da33f9fa2f85 take care of that. Unfortunately I committed them to the wrong branch :sweat_smile:

hryniuk: Oh, didn't know. Wouldn't it be better to express that with proper enum? Something like:
```cpp
enum class Corners
{
    Triangle = 3,
    Square = 4,
    Hexagon = 6
};
```

hryniuk: Here's the pull request with it: https://github.com/serin-delaunay/Wangscape/pull/33, if you like it.

Issue #38, opened 2016-12-19
Make borders agree in Perlin tiles


serin-delaunay:
The Perlin tileset generation algorithm is outputting tiles that resemble what was intended, but vertical and horizontal borders are not matching up. This is visible in both soft and sharp versions of alpha mask generation:

![perlin-hard](https://cloud.githubusercontent.com/assets/15715657/21314770/b9979930-c5f0-11e6-980b-a73486a96a5c.PNG)
![perlin1](https://cloud.githubusercontent.com/assets/15715657/21314772/bc9891c0-c5f0-11e6-91ba-6bc3568678fe.PNG)

serin-delaunay:
Resolved in https://github.com/serin-delaunay/Wangscape/commit/5a3ff7867172b167ae811ccd66a23bb5d805544a. This issue was never present in `master`, so there's no need to wait for #50 to be merged before closing it.

Issue #37, opened 2016-12-19
More options for post-processing corner weight masks (perlin branch)


serin-delaunay:
The corner weight masks output by the `TilePartitionerPerlin` class are floats, in theory nonnegative (though this is not guaranteed) and there is no constraint on their total weight.

The alpha masks used to write tiles are `UInt8s`, and must sum to 255 in non-error situations.

`TilePartitionerBase::applyWeights` is currently used to convert corner weight masks to alpha masks. So far two algorithms have been used:
* One that tries to assign the 255 alpha points with a distribution as close as possible to the relative weights in the corner masks.
* One that assigns all 255 alpha points to the corner whose mask has the greatest weight in that pixel.

`applyWeights` should be separated from the `TilePartitioner` class into either a set of classes with a common base, or a set of functions with a common signature. The resulting code should be able to apply any reasonable combination of these effects:

- Local rules:
  - [ ] Distribute alpha points according to relative weights.
  - [ ] Distribute alpha points according to relative weights raised to power p (the higher the value of p, the sharper the transitions).
  - [ ] Distribute all alpha points to the terrain with the highest weight.
- Global rules:
  - [ ] Some terrain types may "win" and suppress other terrain types in a pixel.
  - [ ] Some terrain types may prevent other terrain types from forming intruding disconnected patches.
  - [ ] Some terrain types may be unable to appear as intruding disconnected patches.
  - [ ] Some terrain type pairs may have a different texture at their borders, and these textures may be subject to the effects above (#14).

Issue #28, opened 2016-12-13
Build ModuleGroups from specs in options file


serin-delaunay:
Wangscape tile generation is of course intended to be customisable. A draft of the flow of data between noise modules is given in #15. Three of the modules are "anything", which means they should be completely specified by the user. All user customisation of Wangscape is mediated by the JSON options file. So, a function must be made that converts a JSON value to a ModuleGroup ([examples](https://github.com/serin-delaunay/Wangscape/blob/perlin/Wangscape/ModuleFactories.cpp)) containing [modules from libnoise](http://libnoise.sourceforge.net/docs/modules.html) (and possibly also custom modules defined in Wangscape).

* ~Write a JSON schema which accepts a JSON value iff it describes a valid ModuleGroup. Features:~
  * ~Each component module given a name which is unique within the ModuleGroup.~
  * ~Module types and parameters all customisable~
  * ~Module parameters restricted to those available in the module type?~
  * ~Module sources either referred to by name, or determined hierarchically, or both (implementor's choice)~
* ~Write a routine which takes a const rapidjson::Value& as input and returns a ModulePtr containing a ModuleGroup built according to the input data.~
* ~Write example JSON documents and tests to check the ModuleGroup is built correctly for valid inputs (validating using the schema may be deferred until #21 is resolved).~

TODO moved to #83.

serin-delaunay:
[RESTnoise](http://restnoise.org) already has a standard for JSON -> libnoise module conversion. An example JSON file is [here](http://restnoise.org:3001/networks/name:world2?gradient=realistic&bound0=-90&bound1=90&bound2=-180&bound3=180&maptype=sphere). It would probably be best to reuse their format. I can't find a JSON schema for RESTnoise; it might not exist yet.

The customisable masks are all intended to be used as squares ([0,1]x[0,1], [0,1]x[-1,0], or [-1,0]x[0,1]), so some of the geometrical features used in RESTnoise will not be needed. This may change if non-square tile geometries are supported in the future.

serin-delaunay:
Some of the options in #37 are less suited as part of the AlphaCalculator step and more suited to being part of the Perlin noise module stack. In particular, applying a function such as raising all weights to the same power should be implemented with libnoise modules, and should be customisable in a similar way.

However, customising these postprocessing steps will have to be implemented differently from customising the source modules, because the postprocessing steps will have an input module.

serin-delaunay:
[...]

Issue #39, opened 2016-12-19
Output full debug tilesets


serin-delaunay:
#38 was an issue caused by incorrect weights in the alpha masks generated for tiles, and it would be best to investigate similar issues by looking at the noise textures contributing to the generated tiles. Since it was only a visible issue when looking at more than one tile, it would be preferable to have visualisations of all the noise textures for all the tiles simultaneously available. The present debug output facilities only make provisions for outputting noise textures for one tile at a time, so the program architecture must be adapted to this requirement.

In future, users customising noise modules will probably also want to see debug tilesets, so this will have long-term benefits as well as being a necessary tool in development.

serin-delaunay:
#38 is fixed already. :tada: But implementing this should still be a high priority.

PR #40, opened 2016-12-20
Perlin tile generation

serin-delaunay:
Changes:
* Adds libnoise dependency.
* Renames TilePartition* to TilePartitioner* and turns them into derived classes.
* Adds wrapper classes for libnoise Modules:
  * ModuleGroup, which allocates `map<shared_ptr<Module>>` storage.
  * Reseedable, which groups a `shared_ptr<Module>` with a function object that reseeds all reseedable Module dependencies, and can build ModuleGroups using operators and functions.
* Adds noise-based TilePartitioner with tests.
* Adds AlphaCalculator class with static functions for converting a pixel's corner weights into alpha values that sum to 255.

Status:
* Builds
* Most tests pass
* Borders don't match in output tiles (#38)
* Noise modules not data-driven (#28)
* Post-processing using AlphaCalculator is not customisable and is missing features (#37)


Wangscape/TilePartitionerBase.h
```diff
@@ -0,0 +1,18 @@
...
+#include <vector>
+#include <SFML/Graphics.hpp>
+#include <functional>
+#include "Options.h"
+
+class TilePartitionerBase
+{
+public:
+    typedef std::vector<std::pair<sf::Texture, TerrainID>> TilePartition;
+    typedef std::vector<TerrainID> Corners;
```
hryniuk: Oh, another `typedef`. There are too many of them for me. It seems for that it will be better to reduce the numbers of `typedef`s and move them to separate files.

I like to treat them like classes (new types) and put one `typedef` in one file (not always, of course). It could seem to you that this way there will too many files with two or three lines, but in my opinion it's just matter of dividing source code into modules. In Wangscape we have roughly (as I noticed):
* Reader
  - JSON options
  - command line arguments
  - input images
* Generator
  - current tileset generator
  - modifiers
    + borders
    + **Perlin noise**
* Writer
  - output images
  - output JSONs

We could (and I think we should) map it to a directory structure and namespaces.

Wangscape/TilePartitionerPerlin.cpp
```diff
@@ -0,0 +1,103 @@
...
+    write_map(cc, "cc", debug_map);
+    write_map(border_xy, "bhv", debug_map);
+    Reseedable max_det = makeLinearMovingScaleBias(makeConst(1.), left, top, 0.85, 0.15);
+    Reseedable min_det = makeLinearMovingScaleBias(makeConst(-1.), left, top, 0.85, 0.15);
+    write_map(min_det, "min_det", debug_map);
+    write_map(max_det, "max_det", debug_map);
+    write_map(deterministic, "det", debug_map);
+    write_map(ef, "ef", debug_map);
+    write_map(corner, "corner", debug_map);
+#endif
```
hryniuk: Move it to a function, please.

hryniuk: Also: we could control it with `-v` command line argument 😃.

Wangscape/TilePartitionerPerlin.cpp
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+    }
+    for (int i = 0; i < 4; i++)
+    {
+        sf::Texture t;
+        t.loadFromImage(outputs[i]);
+        regions.push_back({t, corners[i]});
+    }
+}
```
hryniuk: Could this function be split?

WangscapeTest/TestTilesetGenerator.cpp
```diff
@@ -19,38 +19,41 @@ class TestTilesetGenerator : public ::testing::Test {
...
+//        EXPECT_STREQ(corner.c_str(), "s") << "Incorrect corner in last tile";
+//    }
+//    for (size_t x = 0; x < 4; x++)
+//    {
+//        for (size_t y = 0; y < 4; y++)
+//        {
+//            EXPECT_FALSE(tiles.find({ x,y }) == tiles.end()) << "Missing tile";
+//        }
+//    }
+//}
```
hryniuk: Is there a need to keep it, even commented out?

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
+#include "AlphaCalculator.h"
+#include <algorithm>
+#include <numeric>
+
+
+
+AlphaCalculator::AlphaCalculator()
```
serin-delaunay: Agreed. This (and others like it) can be removed.

Visual Studio automatically makes constructors when creating a class; I should look for a way to disable that feature.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+}
+
+
+AlphaCalculator::~AlphaCalculator()
+{
+}
+
+void AlphaCalculator::calculatePixelAlphaLinear(WeightVector& weights, AlphaVector& alphas)
```
serin-delaunay: I agree, this isn't the ideal way to store this data or transform it. In future there will be more stages of transformation between noise evaluation and tile generation (#37), and more textures to be blended. So the way data is passed through this function will probably change significantly.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <numeric>
+
+
+
+AlphaCalculator::AlphaCalculator()
+{
+}
+
+
+AlphaCalculator::~AlphaCalculator()
```
serin-delaunay: Thanks for pointing that out.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+}
+
+void AlphaCalculator::calculatePixelAlphaLinear(WeightVector& weights, AlphaVector& alphas)
+{
+    int err = 255;
```
serin-delaunay: You're right, something like `remainder` would be a more accurate description. It's the number of alpha points which still need to be assigned to any of the textures.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+    float total_weight = 0.f;
+    for (auto& weight : weights)
+    {
+        weight = std::max(0.f, weight);
```
serin-delaunay: I accidentally passed negative weights to this function while developing this process, and it's inevitable that end-users will too, so the weight-to alpha conversion needs to be able to cope with negative inputs without crashing.

But you're right that applying a lower bound should not be this function's job - that should be split into a separate function.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+        total_weight += weight;
```
serin-delaunay: Yes, that would be a better way to write that step.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+        // where this does happen.
+        for (auto& alpha : alphas)
+        {
+            alpha = 0;
+        }
+        return;
+    }
+    for (size_t i = 0; i < weights.size(); i++)
+    {
+        int alpha = (int)((255 * weights[i]) / total_weight);
```
serin-delaunay: `total_weight <= 0.f` will catch `total_weight == 0`, but there is still danger of overflow when it is positive but very small. The comparison should be made to a very small number (1000*epsilon would probably be enough) to catch that case.

serin-delaunay: I agree there should be more assertions made, but I don't think this is a suitable site for them. Most of the calculations performed by Wangscape will eventually become user-configurable, and when they go wrong, it will be important to show the context of the incorrect calculations to the user. The best way to do that won't be to terminate with an exception, but to output debug tilesets showing the important steps of the calculation. Writing a message in the log file would also be a good idea.

Wangscape/NoiseMap.h
```diff
@@ -0,0 +1,66 @@
+#pragma once
+#include <vector>
+#include <noise/noise.h>
+#include <SFML/Graphics.hpp>
+#include <functional>
+
+class NoiseMapBase
```
serin-delaunay: Yes, that would be best.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void AlphaCalculator::calculatePixelAlphaMax(WeightVector& weights, AlphaVector& alphas)
+{
+    for (auto& alpha : alphas)
+    {
+        alpha = 0;
+    }
+    auto max_it = std::max_element(weights.cbegin(), weights.cend());
+    auto max_idx = std::distance(weights.cbegin(), max_it);
```
serin-delaunay: You're right, `idx` is not much of an abbreviation and with `max_` it's even less worthwhile.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+            alpha++;
+            err--;
+        }
+        i++;
+    }
+}
+
+void AlphaCalculator::calculatePixelAlphaMax(WeightVector& weights, AlphaVector& alphas)
+{
+    for (auto& alpha : alphas)
```
serin-delaunay: Yes, that should be factored out.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+    *alpha_it = 255;
+}
+
+void AlphaCalculator::calculatePixelAlphaFunction(WeightVector& weights, AlphaVector& alphas,
+                                                  std::function<float(float)> fn)
+{
+    for (auto& weight : weights)
+    {
+        weight = fn(weight);
+    }
```
serin-delaunay: I hadn't come across std::transform before, thanks.

Wangscape/AlphaCalculator.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <vector>
+#include <functional>
+#include <SFML/Config.hpp>
+class AlphaCalculator
```
serin-delaunay: Agreed. Many other files could also benefit from better ordering of `#include`s.

Wangscape/MakeReseedable.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+#include <noise/noise.h>
+#include "Reseedable.h"
+#include "ModuleGroup.h"
+
+void dontSetSeed(int seed);
+
+// http://stackoverflow.com/a/9531274
+// Fails in VS2015
```
serin-delaunay: VS2017 seems to have better SFINAE support, and also seems to be freely available, so I might try it out later. Then perhaps this can be replaced with a more DRY solution.

Wangscape/NoiseMap.h
serin-delaunay: Yes, good idea!

WangscapeTest/TestTilesetGenerator.cpp
serin-delaunay: IIRC they're temporarily commented out because they fail to compile after the changes in this branch; it would be better to rewrite them and restore the functionality.

Wangscape/TilePartitionerBase.h
serin-delaunay: Good idea. The codebase is already large enough that applying a directory structure would be appropriate, and many of the typedefs don't really belong inside a class.

Wangscape/AlphaCalculator.h
```diff
@@ -0,0 +1,20 @@
...
+{
+public:
+    typedef std::vector<float> WeightVector;
+    typedef std::vector<sf::Uint8> AlphaVector;
```
serin-delaunay: Yes, `using` statements for some of the most common names from `std` would be reasonable.

Wangscape/AlphaCalculator.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+AlphaCalculator::PixelAlphaCalculator AlphaCalculator::makeCalculatePixelAlphaPower(float power)
+{
+    return bindCalculatePixelAlphaFunction(std::bind(powf, std::placeholders::_1, power));
+}
+
+AlphaCalculator::PixelAlphaCalculator AlphaCalculator::bindCalculatePixelAlphaFunction(std::function<float(float)> fn)
+{
+    return std::bind(calculatePixelAlphaFunction, std::placeholders::_1, std::placeholders::_2, fn);
```
serin-delaunay: I don't think they're very important right now. When #37 is implemented (it's not my current task) they will probably either disappear or be used more extensively.

Applying float->float and float->alpha transformations will certainly be separated and modularised (here they're applied in one function).

Wangscape/TilePartitionerPerlin.cpp
serin-delaunay: Yes, my current task is to replace these lines with a more complete and flexible system for outputting debug tilesets. The user should be able to switch debug output on/off for each module, and debug output should be for full tilesets rather than single tiles.

In the meantime, these lines would be better in a function, and controlled by a command line flag rather than a compiler flag.

Wangscape/AlphaCalculator.h
```diff
@@ -0,0 +1,20 @@
...
+    typedef std::function<void(WeightVector&, AlphaVector&)> PixelAlphaCalculator;
+    AlphaCalculator();
+    ~AlphaCalculator();
+    static void calculatePixelAlphaLinear(WeightVector& weights, AlphaVector& alphas);
+    static void calculatePixelAlphaMax(WeightVector& weights, AlphaVector& alphas);
+    static void calculatePixelAlphaFunction(WeightVector& weights, AlphaVector& alphas,
+                                            std::function<float(float)> fn);
+    static PixelAlphaCalculator makeCalculatePixelAlphaPower(float power);
+    static PixelAlphaCalculator bindCalculatePixelAlphaFunction(std::function<float(float)> fn);
```
serin-delaunay: Yes, `AlphaCalculator` has no reason to be a class at the moment. In future I would like to make these calculations composable like `noise::Module`s. Many of these operations (applying lower bound, applying power) should probably be moved into the noise module stack.

Wangscape/Reseedable.cpp
```diff
@@ -0,0 +1,285 @@
...
+    std::shared_ptr<noise::module::Select> select_p = std::make_shared<noise::module::Select>();
+    select_p->SetSourceModule(0, *source_a.module);
+    select_p->SetSourceModule(1, *source_b.module);
+    select_p->SetControlModule(*module);
+
+    std::shared_ptr<ModuleGroup> mg = std::make_shared<ModuleGroup>();
+    mg->insert("control", *this);
+    mg->insert("source0", source_a);
+    mg->insert("source1", source_b);
+    mg->insert("output", makeReseedable(select_p));
```
serin-delaunay: Yes, good idea. That way a more expressive name can be given to `mg`.

Wangscape/CornerCombinerBase.h
```diff
@@ -0,0 +1,34 @@
...
+{
+public:
+    CornerCombinerBase(double power = 1.);
+    virtual int GetSourceModuleCount() const
+    {
+        return 0;
+    }
+
+    virtual double GetValue(double x, double y, double z) const;
+    double power;
```
serin-delaunay: I don't think it needs to be private; nothing extra needs to be done when setting the value.

Wangscape/CornerCombinerBase.h
```diff
@@ -0,0 +1,34 @@
...
+// when x, y are in [-1., 1.].
+//
+// When power is greater than 1.0, the magnitude of
+// the output will be greater than 1.0 along the axes
+// when x, y are in [-1., 1.], and the output should
+// be clamped.
+class CornerCombinerBase : public noise::module::Module
+{
+public:
+    CornerCombinerBase(double power = 1.);
```
serin-delaunay: Yes!

Wangscape/Exp.h
```diff
@@ -0,0 +1,77 @@
...
+{
+public:
+    double base;
+    Exp() :
+        Module(GetSourceModuleCount()),
+        mBase(2.)
+    {};
+    virtual int GetSourceModuleCount() const
+    {
+        return 1;
```
serin-delaunay: This is a quirk of the libnoise library. It uses this number to allocate an array of `noise::Module*`s used as input values. For instance, a `Checkerboard` module has 0 inputs, an `Abs` module has 1 input, an `Add` module has 2.

We do need to specify this number manually for each class derived from `Module`, but we could add some intermediate abstract classes with 0, 1, 2... input modules. Then the custom `Module`s could be derived from those with less verbosity.

Wangscape/ModuleFactories.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <stdexcept>
+
+using noise::module::Abs;
+using noise::module::ScaleBias;
+using noise::module::Clamp;
+using noise::module::TranslatePoint;
+using noise::module::ScalePoint;
+using noise::module::Perlin;
+
+Reseedable makeQuadrantSelector(Reseedable& source, bool x, bool y)
```
serin-delaunay: Yes, `x_positive` and `y_positive` would be better here.

Some of the `int`/`size_t` `x` and `y` variables should also maybe be renamed to clarify whether they are coordinates in a tile bitmap, a tileset bitmap, or tile indices in a tileset. The `float`/`double` `x` and `y` variables are probably OK as they are.

Wangscape/Reseedable.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return makeReseedable(mg);
+}
+
+Reseedable Reseedable::rotatePoint(double x_angle, double y_angle, double z_angle)
+{
+    std::shared_ptr<noise::module::RotatePoint> rotate_point_p = std::make_shared<noise::module::RotatePoint>();
+    rotate_point_p->SetSourceModule(0, *module);
+    rotate_point_p->SetXAngle(x_angle);
+    rotate_point_p->SetYAngle(y_angle);
+    rotate_point_p->SetZAngle(z_angle);
```
serin-delaunay: `noise::RotatePoint` doesn't support chaining. But using `SetAngles()` would make the function more concise.

Wangscape/Reseedable.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+    std::shared_ptr<ModuleGroup> mg = std::make_shared<ModuleGroup>();
+    mg->insert("source", *this);
+    mg->insert("output", makeReseedable(translate_point_p));
+    return makeReseedable(mg);
+}
+
+Reseedable Reseedable::turbulence()
+{
+    throw;
```
serin-delaunay: Better to just implement the function :) but in future I'll use `std::terminate` or `throw` a real exception when marking an unimplemented function, rather than `throw` nothing.

Wangscape/Reseedable.cpp
```diff
@@ -0,0 +1,285 @@
...
+#include "Exp.h"
+
+double Reseedable::getValue(double x, double y, double z)
+{
+    return module->GetValue(x, y, z);
+}
+
+Reseedable Reseedable::abs()
+{
+    std::shared_ptr<noise::module::Abs> abs_p = std::make_shared<noise::module::Abs>();
```
serin-delaunay: Yes, `auto` would be appropriate here.

`abs_p` can't be `const` because:
* `abs_p->SetSourceModule(0, *module);` is a non-const function call.
* `abs_p` will be stored in a `Reseedable`, and `Reseedable` stores a non-const `ModulePtr` because the `SetSeed` function may alter that `ModulePtr`.

In this case, `abs_p` does not have a seed to set, so the `SetSeed` function will be a no-op. But the current version of the `Reseedable`/`ModuleGroup` system can't accommodate a const `ModulePtr`.

However, this system is quite fragile as it stands. `Reseedable` and `ModuleGroup` data members are public and type information is not included. This means one could write code that alters the network of noise modules after construction, but it would need type information stored externally (or from RTTI), and would almost certainly have errors.

My first instinct would be to refactor the code with as much as possible `private`/`protected`/`const`. However, when implementing #17, that would probably mean that the whole network needs to be rebuilt when altered. That might not be a bad thing.

Wangscape/NormLPQ.cpp
```diff
@@ -0,0 +1,34 @@
...
+    Module(GetSourceModuleCount())
+{
+    setPQ(p, q);
+}
+
+double NormLPQ::GetValue(double x, double y, double z) const
+{
+    double ax = std::abs(x);
+    double ay = std::abs(y);
+    double az = std::abs(z);
```
serin-delaunay: I don't think so. Libnoise is explicitly built to work with 3D noise, and even if we altered the API of the Wangscape `Module`s to accept a struct, the libnoise `Module`s would still be used through `GetValue(x,y,z)`.

We could start using `noise::model::Plane` to remove the `z` from some of the calls. The `Module`s would still need to accept `z` as an argument.

Wangscape/NoiseModuleManager.cpp
```diff
@@ -0,0 +1,53 @@
...
+        return r;
+    return r.translatePoint(-1., 0., 0.);
+}
+
+Reseedable NoiseModuleManager::getBorderHorizontal(TerrainID left, TerrainID right, bool y_positive)
+{
+    Reseedable& r = mBordersHorizontal.at({left, right});
+    if (y_positive)
+        return r;
+    return r.translatePoint(0., -1., 0.);
```
serin-delaunay: This is what `makeQuadrantSelector` is for. I should use that here (and in any other lines that do this kind of work).

Wangscape/TileGenerator.cpp
```diff
@@ -33,7 +32,7 @@ void TileGenerator::generate(sf::RenderTexture& image, size_t x, size_t y,
         sf::Sprite tile(temp.getTexture());
         tile.setPosition(sf::Vector2f((float)x, (float)y)*(float)options.resolution);
         image.draw(tile, sf::RenderStates(sf::BlendMode(sf::BlendMode::SrcAlpha, sf::BlendMode::One, sf::BlendMode::Add,
-                                                        sf::BlendMode::Zero, sf::BlendMode::One, sf::BlendMode::Add)));
+                                                        sf::BlendMode::Zero,     sf::BlendMode::One, sf::BlendMode::Add)));
```
serin-delaunay: The extra 4 spaces are there to align the two lines. Arguments 1-3 and arguments 4-6 are symmetric in this constructor for `sf::BlendMode`. ([link](http://www.sfml-dev.org/documentation/2.4.1/structsf_1_1BlendMode.php#a69a12c596114e77126616e7e0f7d798b))

Wangscape/AlphaCalculator.cpp
hryniuk: Right, I missed a `return`.

Wangscape/Exp.h
hryniuk: Oh, so I think setting it manually (leaving as it is) is a better solution. They could be consts.

Wangscape/TileGenerator.cpp
hryniuk: Ok, ok 😄.

Wangscape/ModuleGroup.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return mModules.at(output_id).module.get()->GetValue(x, y, z);
+}
+
+void ModuleGroup::SetSeed(int seed)
+{
+    for (auto it : mModules)
+    {
+        // Use a different seed for each submodule
+        int hash = (int)std::hash<ModuleID>{}(it.first);
+        it.second.setSeed(seed^hash);
```
serin-delaunay: I don't think `std::transform` is appropriate here. This loop doesn't assign to the elements of `mModules`, instead it calls each element's `setSeed` function.

Wangscape/NoiseModuleManager.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "ModuleFactories.h"
+#include <random>
+#include <time.h>
+
+NoiseModuleManager::NoiseModuleManager(const Options & options) :
+    mRNG((unsigned int)time(nullptr))
+{
+    for (const auto& terrain : options.terrains)
+    {
+        mStochasticMasks.insert({terrain.first, makePlaceholder().scaleBias(0.5, 0.5)});
```
serin-delaunay: This loop uses the keys from `options.terrains`, but `std::transform` on a `std::map` seems to give access only to the values. I don't think `std::transform` can be used here unless the structure of `options.terrains` changes.

Wangscape/ModuleGroup.cpp
hryniuk: Right, I probably thought about `std::for_each` or sth, but this for is enough.

Wangscape/Exp.h
```diff
@@ -0,0 +1,77 @@
...
+protected:
+    double mBase;
+};
+
+// Noise module that raises a source module to an exponent.
+// No attempt is made to guard against invalid operations.
+// Make sure the range of the source module
+// is compatible with the exponent!
+// The formula is pow(source_value, exponent)
+class Power : public noise::module::Module
```
serin-delaunay: Done (and renamed to avoid conflict with noise::module::Power).

Wangscape/NoiseMap.h
serin-delaunay: Moved to subdirectories and namespaces. Still need to remove the prefix.

Wangscape/TilePartitionerPerlin.cpp
serin-delaunay: Actually, These lines should just be removed; there aren't currently any bugs that these lines of code can help to resolve. Those that remain will need #39, and #39 will need a different method of output.

serin-delaunay:
Transferring the repository to the organisation broke my ability to add commits to this PR, so I've moved it to #50.

PR #45, opened 2016-12-27
Port from RapidJSON to spotify-json

hryniuk:
Hello, finally, I got it. I removed RapidJSON dependency and simplify a bit both source and test code. I refactored the code a bit, but still a few things should be made differently IMO... but that's a task for the future, we've got a lot of issues to do.

I've changed a bit the output JSON of a tile data, so I suppose some change should be done in **Wangview**, I can do it in the next PR.

It builds, works on examples and tests pass.

Hope you like it, waiting for review :wink:. After merge, we'll be able to close Travis PR 🎆!


Wangscape/Options.cpp
```diff
@@ -1,94 +1,5 @@
...
-            {
-                c.push_back(jt.GetString());
-            }
-            cliques.push_back(c);
-        }
-    }
-}
 
-Options::~Options()
+Options::Options(std::string /*filename*/)
```
hryniuk: Yes, I'll remove it.

CMakeLists.txt
```diff
@@ -6,5 +6,6 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
+add_subdirectory(lib/spotify-json)
```
hryniuk: Ok, I'll check it.

Wangscape/CMakeLists.txt
```diff
@@ -1,17 +1,26 @@
 set(lib-src
     MetaOutput.cpp
     Options.cpp
+    OptionsManager.cpp
```
hryniuk: Yes, there's a thing called `GLOB`, which allows to include files that conform to the given pattern. I thought that both ways to do it are equally good, but I've found [this answer](http://stackoverflow.com/a/32412044/4169253) which convice me to list files explicitly.

Now we've got one library built from source files, when we got the second, with such list it will be easier to say, which files every library is built from.

Wangscape/MetaOutput.h
```diff
@@ -21,14 +24,16 @@ class MetaOutput
     void writeTilesetData(std::string filename) const;
     void writeTerrainHypergraph(std::string filename) const;
     void writeAll(const Options& options) const;
-    const rapidjson::Document& getTileData() const;
-    const rapidjson::Document& getTileGroups() const;
-    const rapidjson::Document& getTilesetData() const;
+    const std::vector<Tile>& getTileData() const;
+    const std::unordered_map<std::string, std::vector<TileLocation>>& getTileGroups() const;
```
hryniuk: Ok, I'll look for such type names and I'll move them to a separate files.

`unordered_map` has a bit better performance, thus, as far as I know, is better for lookup usage. The difference is that `unordered_map` needs hash function for custom keys, but that's not the case. I think that's additional argument for the typename.

Wangscape/OutputFilenames.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <string>
+
+struct OutputFilenames
```
serin-delaunay: (Or `metaoutput::Filenames`)

Wangscape/OptionsManager.cpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    std::string str{std::istreambuf_iterator<char>(ifs),
+                    std::istreambuf_iterator<char>()};
+
+    mOptions = spotify::json::decode<Options>(str.c_str());
+    mOptions.filename = filename;
+
+    // TODO(hryniuk): move it elsewhere
+    auto outputDirectory = mOptions.outputDirectory;
```
hryniuk: Right 😄.

Wangscape/OptionsManager.cpp
```diff
@@ -0,0 +1,54 @@
...
+    std::ifstream ifs(filename);
+    if (!ifs.good())
+    {
+        throw std::runtime_error("Could not open options file");
+    }
+
+    std::string str{std::istreambuf_iterator<char>(ifs),
+                    std::istreambuf_iterator<char>()};
+
+    Options options = spotify::json::decode<Options>(str.c_str());
```
hryniuk: Oh, it should be `mOptions = ...`.

Wangscape/Tile.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+struct Tile
```
hryniuk: If you're not against extending this PR, I'll create this namespace and directory and I'll move file connected to `MetaOutput` to it.

lib/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(spotify-json)
```
hryniuk: I've got to check it. I don't know now, why it's repeated.

Wangscape/Tile.h
serin-delaunay: A namespace and directory would be great :)

CMakeLists.txt
hryniuk: There is the `EXCLUDE_FROM_ALL` flag for `add_subdirectory` to disable including subdirectory targets to the project's targets list.

serin-delaunay: That sounds like what we need.

Wangscape/OptionsManager.cpp
hryniuk: No! They should be in a separate function for now :).

I'm sure that we should move an output directory creation to a different component in the end, but it's not a thing to do now. It's really nasty that during input JSON reading user gets an error about an output directory, even before Wangscape starts looking for any image.

serin-delaunay: Hmm, yes, you're right. Creating the output directory shouldnt be part of `loadOptions`, or `OptionsManager`. But if creating that directory fails, the error should be detected before any images are loaded or generated. The exception would be in eg. test cases where options must be loaded but output will not be written.

In any case, the `loadOptions` code shouldn't be duplicated anywhere.

lib/CMakeLists.txt
hryniuk: I've changed `Wangscape/CMakeLists.txt` to include only `lib/`.

Wangscape/OptionsManager.cpp
hryniuk: Wangscape could always save the output in a current directory or ask for an another path. I don't understand why it cannot load and generate any images if it's able to do so. Wangscape uses SFML, so it could at least display generated images, if it cannot save it in a given directory.

Moreover, we could make it a feature (testing different parameters will be easier, I think) :wink:.

serin-delaunay: Wangscape can generate a lot of output files and overwrite existing files, so it shouldn't be allowed to automatically choose an unexpected directory when it fails to open the user's choice of directory.

Prompting the user for an alternative output dir would be OK (but maybe unnecessary, if the fault is detected quickly).

Using SFML to display previews during or after tileset generation is a nice idea for the future.

hryniuk:
Ok, I've moved MetaOutput to subdirectory and a namespace ([this commit](https://github.com/Wangscape/Wangscape/pull/45/commits/f32c05065595332b08fbb55c43c3550d26b68f72)) :).

By the way I've fixed namings and reorganized libraries.

serin-delaunay:
Great, this is almost ready to merge. We just need:
* `TileGroups` typedef
* Updated `README.md`
* I need to make sure it builds on Windows

hryniuk:
Done.

serin-delaunay:
I'll need to make some changes to get it building in VS (project file changes, https://github.com/google/double-conversion in a submodule), but I think that'll be simpler to do in a separate pull request. Merged :)

hryniuk:
Great! Next PR will be much smaller (I hope yours too :wink:).

Issue #44, opened 2016-12-27
Tests parameters are hard-coded, in different files


hryniuk:
It should be:
* common test fixture
* config file (path to it passed as command line argument)
* command line argument (a path to `example_options.json`)

serin-delaunay:
Test parameters should definitely be made more flexible. I would prefer not to have obligatory command-line parameters in the test executable, so there should be a test config file with the path to `example_options.json`, a hardcoded default location for it, and an optional command line argument to specify a different one.

hryniuk:
We could move this test options to `WangscapeTest` directory to state explicitly that this file is a part of tests.

serin-delaunay:
Yes, the test config file should definitely be in `WangscapeTest` folder.

The examples shouldn't be in the test folder; they're meant not just as test data, but also as demonstrations and templates for end users to build on.

hryniuk:
I think that examples shouldn't be in a source code directory as well.

So... maybe the problem is just a current project structure? It will be a bit better to have some "standard structure". We could move current content of the `Wangscape` dir up and have `examples` under `doc` dir e.g.:
```
|- doc/
    |- examples/
        |- example_1
        |- example_2
        |- ...
        |- example_n
|- lib/
|- src/
|- test/ [or tst/]
```

It's just a suggestion, it could be next PR to merge (with a little discussion before) after those three that we've got now :wink:.

serin-delaunay:
Yes, `doc/examples/` is probably the best place fot the examples. `test/` will be fine for the tests.

serin-delaunay:
I think this is fixed now. Restructuring the project folder structure is really a different issue.

PR #47, opened 2016-12-31
Visual Studio + spotify-json

serin-delaunay:
This PR includes the code necessary for a non-CMake VS build.

I updated several `#include` directives which didn't agree with the new directory structure.

I've made a batch file to invoke CMake for spotify-json and double-conversion. Spotify-json's CMakeLists.txt forces the compiler switch /MT[d], which makes it incompatible with the Google Test NuGet package (compiled with /MD[d]).

The two options to resolve this issue were to add GTest as a submodule, or to alter spotify-json's build script to use /MD[d] (I couldn't figure out how to do that on the CMake command line). For now, I chose the latter. This can be changed if necessary. Considering that GTest must be built from source on Travis, it might be preferable to use a submodule. Otherwise, the spotify-json submodule should probably point to a fork with the necessary changes.

I changed the example paths in the test files back to one level below the project root. If it's important on Linux to run the test suite from two levels below the project root, then I can instead change the working directory I run the tests with. After #44 is resolved I expect this won't be a problem.


serin-delaunay:
#50 did this branch's job.

PR #50, opened 2016-12-31
Perlin 2

serin-delaunay:
Transferring the main repository to the organisation account seems to have broken my ability to add commits to https://github.com/Wangscape/Wangscape/pull/40. I've no idea how to fix that, and it wasn't ready to merge so I'm making a new PR.

#39 is unresolved. This PR is already very large and resolving it will require even more structural changes. It's probably better to do it in a separate PR, after this is merged.

Changes still required (https://github.com/Wangscape/Wangscape/pull/40#pullrequestreview-14364296):
- [x] Refactor `AlphaCalculator`:
  - [x] Pass and return values in a more sensible way.
  - [x] Extract corner-independent operations and preconditions on weights (raising to a power, setting a lower bound, zeroing alpha values).
  - [x] Calculate `total_weight` with `std::accumulate`.
  - [x] Prevent float overflow when `total_weight` is very small.
  - [x] Rename `AlphaVector`, `WeightVector`, etc.
  - [x] Simplify `AlphaCalculator`; when implementing https://github.com/Wangscape/Wangscape/issues/37 it will need to be a class hierarchy for the sake of composability, but with the other changes, only calculatePixelAlphaLinear and calculatePixelAlphaMax will need to be preserved.
- [x] Some constructors and destructors probably still need to be `default` and `virtual`.
- [x] Many constructors should be marked `explicit`.
- [x] Try to split `TilePartitionerPerlin::makePartition` more.
- [x] Try to fix `TestTilesetGenerator::TestGenerateClique`, or remove it.
- [x] Fix the two failing tests in `TestMovingScaleBias`.
- [x] More of the source should be divided into folders. ~I think that's an issue for another PR, though.~
- [ ] `makeReseedable` should eventually be made generic using  SFINAE, but it will require a change of compiler for me (and I don't think it's urgent). Probably an issue for another PR.
- [ ] Resolve merge conflicts.


hryniuk:
I see that rebasing on master is needed and most likely it will be painful :worried:.

**Good luck** 🍀 **&** 🎆 **Happy New Year** 🍾

serin-delaunay:
Happy Prime Year!

serin-delaunay:
OK, I think this is ready for final review and conflict resolution.

serin-delaunay:
(If you'd prefer to avoid the rebasing, I can try merging Wangscape/master into this branch to resolve the conflicts)

Wangscape/noise/Reseedable.h
```diff
@@ -0,0 +1,50 @@
...
+
+    Reseedable abs();
+    Reseedable clamp(double lower, double upper);
+    Reseedable pow(Reseedable& exponent);
+    Reseedable pow(double exponent);
+    Reseedable exp(double base);
+    Reseedable invert();
+    Reseedable scaleBias(double scale, double bias);
+    //Reseedable curve(...?)
+    //Reseedable terrace(...?)
```
hryniuk: Is there a need to keep it commented out :wink:? It could be an issue, if you anticipate necessity of implementing these methods.

Wangscape/noise/module/ModuleFactories.cpp
```diff
@@ -0,0 +1,109 @@
+#include "ModuleFactories.h"
+#include "Gradient.h"
+#include "NormLPQ.h"
+#include "../MakeReseedable.h"
```
hryniuk: Could you use full paths in include? It helps to determine, which file do you include (e.g if you add `Wangscape/Gradient.h` and remove `Wangscape/noise/module/Gradient.h` it could still work (as root dir is added to include directories in `CMakeLists.txt`).

Wangscape/noise/module/ModuleFactories.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <stdexcept>
+
+namespace noise
+{
+namespace module
+{
+
+Reseedable makeQuadrantSelector(Reseedable& source, bool x_positive, bool y_positive)
+{
+    if (x_positive && y_positive)
```
hryniuk: This `if` is redundant and a bit misleading. Without it this method will work the same way, if I see correctly.

Wangscape/noise/module/ModuleFactories.h
```diff
@@ -0,0 +1,58 @@
...
+// Makes a placeholder noise module (a Perlin noise module with typical parameters).
+Reseedable makePlaceholder(int seed = 0,
+                           int octaves = 5,
+                           double frequency = 1.,
+                           double lacunarity = 2.,
+                           double persistence = 0.5,
+                           NoiseQuality quality = NoiseQuality::QUALITY_FAST,
+                           double origin_x=1.1,
+                           double origin_y=2.2,
+                           double origin_z=3.3);
```
hryniuk: As default values allow to specify only `n` first arguments, maybe it will be better to create a struct (with default values) for these parameters? With chainable methods like `setFrequency(Real frequency)` it will be easier to set desired values for them.

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -1,4 +1,4 @@
-#include "Corners.h"
+#include "../Corners.h"
```
hryniuk: `"tilegen/Corners.h"`

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -7,10 +7,15 @@
 #include <boost/filesystem.hpp>
 #include <boost/iterator/zip_iterator.hpp>
 #include <boost/range.hpp>
-#include "common.h"
+#include "../common.h"
```
hryniuk: `"tilegen/common.h"`

Wangscape/tilegen/TilesetGenerator.h
```diff
@@ -1,24 +1,32 @@
 #pragma once
 #include <SFML/Graphics.hpp>
 #include <memory>
-#include "Options.h"
+#include "../Options.h"
```
hryniuk: `"tilegen/Options.h"`

Wangscape/tilegen/partition/TilePartitionerGradient.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+    std::vector<sf::Image> masks;
+    for (auto corner : corners)
+    {
+        sf::Image img;
+        img.create(mOptions.resolution, mOptions.resolution);
+        masks.push_back(img);
+    }
+    std::vector<float> weights(corners.size(), 0.f);
+    int resolution_sub_1 = mOptions.resolution - 1;
```
hryniuk: `const`

Wangscape/tilegen/partition/TilePartitionerPerlin.cpp
```diff
@@ -0,0 +1,104 @@
+#include "TilePartitionerPerlin.h"
+#include "../../noise/module/ModuleFactories.h"
+#include "../../noise/RasterValues.h"
+#include "../alpha/AlphaCalculatorMax.h"
+#include "../../noise/module/Pow.h"
```
hryniuk: Full paths.

Wangscape/tilegen/partition/TilePartitionerPerlin.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+#include "TilePartitionerBase.h"
+#include <noise/noise.h>
+#include <SFML/Graphics.hpp>
+#include "../../Options.h"
+#include "../../noise/ModuleManager.h"
+#include "../../noise/RasterValues.h"
```
hryniuk: Full paths.

Wangscape/noise/module/ModuleFactories.cpp
serin-delaunay: The method won't work the same way; `source.translatePoint` wraps the `source` module in a `TranslatePoint` module, which adds an extra step to each `GetValue` call. If the `TranslatePoint` module doesn't do anything, then it's not necessary to wrap the module.

hryniuk: Ok, understood.

Wangscape/noise/module/ModuleFactories.h
serin-delaunay: This is a placeholder factory; it's only in use while #28 is unimplemented, and it's not necessary to customise its parameters. If the API needs cleaning up, it would be better to just remove these parameters.

Wangscape/noise/module/ModuleFactories.cpp
serin-delaunay: Done.

Wangscape/tilegen/TilesetGenerator.h
serin-delaunay: These 3 are in the project root, not `tilegen`; is this a typo?

Wangscape/noise/RasterBase.h
```diff
@@ -0,0 +1,33 @@
...
+{
+
+class RasterBase
+{
+public:
+    typedef module::Module NoiseModule;
+    typedef double Real;
+    typedef sf::Rect<Real> Bounds;
+
+    RasterBase() = default;
```
serin-delaunay: No. Not in an abstract class.

Wangscape/noise/RasterBase.h
```diff
@@ -0,0 +1,33 @@
...
+#include <SFML/Graphics.hpp>
+#include <functional>
+
+namespace noise
+{
+
+class RasterBase
+{
+public:
+    typedef module::Module NoiseModule;
```
serin-delaunay: I think the `NoiseModule` typedef can just be removed. It was more useful when `Raster*`/`NoiseMap*` were outside of the `noise` namespace and had to use the typename `noise::module::Module`.

Wangscape/tilegen/TilesetGenerator.h
hryniuk: Oh, yes, sorry. I looked at `..` and I thought that you're going up to some `Wangscape` subdirectory.

So leave `"Options.h"` here, if you agree with me about include paths - I think that's enough to add `Wangscape` subdirectory as include directories and use relative paths starting from there. It makes clear, which file is included.

I've checked it and I made a little "design mistake" with `CMakeLists.txt`: currently `Wangscape/metaoutput/CMakeLists.txt` has
```
target_include_directories(metaoutput PUBLIC ${PROJECT_SOURCE_DIR}/Wangscape)
```
line, which implicitly adds `Wangscape` to include directories of `Wangscape` target.

I think that should be changed (I mean adding 
```
target_include_directories(Wangscape PUBLIC ${PROJECT_SOURCE_DIR}/Wangscape)
```
(could be probably `PRIVATE`) line to the main `CMakeLists.txt`) to make it obvious.

Wangscape/noise/RasterValues.h
```diff
@@ -0,0 +1,31 @@
...
+#include <noise/noise.h>
+#include <SFML/Graphics.hpp>
+#include <functional>
+#include "RasterBase.h"
+
+namespace noise
+{
+
+template <typename T>
+class RasterValues : public RasterBase
```
serin-delaunay: Done using `static_assert`.

Wangscape/noise/RasterImage.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    return mImage.getPixel(x, y);
+}
+
+RasterImage::Colour RasterImage::makeColour(Real value)
+{
+    return Colour(std::min(255, std::max(0, int(-value * 256))),
+                  0,
+                  std::min(255, std::max(0, int(value * 256))),
+                  255);
```
serin-delaunay: I've added a comment in the header file. It might behave differently in derived classes, so its name can't be too specific.

Wangscape/noise/ModuleManager.h
```diff
@@ -0,0 +1,50 @@
...
+public:
+    typedef noise::module::Module Module;
+    typedef std::pair<TerrainID, TerrainID> TerrainIDPair;
+    explicit ModuleManager(const Options& options);
+    virtual ~ModuleManager() = default;
+
+    Reseedable getBorderVertical(TerrainID top, TerrainID bottom, bool x_positive);
+    Reseedable getBorderHorizontal(TerrainID left, TerrainID right, bool y_positive);
+    Reseedable& getStochastic(TerrainID terrain);
+protected:
```
serin-delaunay: Done.

serin-delaunay:
OK, Ready to merge. Most of the changes I made to include paths in #47 will be unnecessary now. One major change you should be aware of is that I'm going to fork the spotify-json repository to fix the compile flags in its `CMakeLists.txt`, and use the fork as a submodule instead of the original repository.

hryniuk:
Nice, have fun 😄. It seems that most conflicts are caused by moving `MetaOutput.*` to a subdirectory, so not a big deal.

That's OK, thanks for info :). After all, it will be still just `git submodule update`.

Wangscape/noise/RasterBase.h
serin-delaunay: ... Actually it is needed, because there are constructors in derived classes that use it. Reinstated.

PR #55, opened 2017-01-03
Command line parameter for test filename

serin-delaunay:
This is a quick and rather dirty solution to #44.
* Path to `example_options.json` is passed as an optional command line parameter
* Filename is detected by raw comparison of `argc` after gtest options are consumed; this can be migrated to boost::program_options if preferred.
* If the filename is not detected, a fallback value is used (starting from the project root).
* The filename is used to initialise an evil static data member of a class.
* The same class is used as a base class for tests which need to access the options filename or parsed file.

I tried to do this using [value-parametrised tests](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#value-parameterized-tests), but initialising the values in `main` doesn't seem to be a well-supported use case.

### DIFF
--- a/WangscapeTest/OptionsFilename.cpp
+++ b/WangscapeTest/OptionsFilename.cpp
@@ -1,0 +1,21 @@
+#include "OptionsFilename.h"
+#include <stdexcept>
+
+namespace
+{
+std::string filename;
+bool filenameSet = false;
+}
+
+const std::string& getOptionsFilename()
+{
+    return filename;
+}
+
+void setOptionsFilename(const std::string & optionsFilename)
+{
+    if (filenameSet)
+        throw std::runtime_error("Options filename already set");
+    filename = optionsFilename;
+    filenameSet = true;
+}
--- a/WangscapeTest/OptionsFilename.cpp
+++ b/WangscapeTest/OptionsFilename.cpp
@@ -15,7 +15,12 @@ const std::string& getOptionsFilename()
 void setOptionsFilename(const std::string & optionsFilename)
 {
     if (filenameSet)
+    {
         throw std::runtime_error("Options filename already set");
-    filename = optionsFilename;
-    filenameSet = true;
+    }
+    else
+    {
+        filename = optionsFilename;
+        filenameSet = true;
+    }
 }
--- a/WangscapeTest/TestMetaOutput.cpp
+++ b/WangscapeTest/TestMetaOutput.cpp
@@ -17,7 +17,6 @@
 class TestMetaOutput : public TestRequiringOptions
 {
 protected:
-    using TestRequiringOptions::TestRequiringOptions;
     tilegen::TilesetGenerator tg;
     const metaoutput::MetaOutput& mo;
     TestMetaOutput() :
--- a/WangscapeTest/TestMetaOutput.cpp
+++ b/WangscapeTest/TestMetaOutput.cpp
@@ -12,17 +12,18 @@
 #include <tilegen/TilesetGenerator.h>
 #include <tilegen/partition/TilePartitionerSquares.h>
 
-class TestMetaOutput : public ::testing::Test {
+#include "TestRequiringOptions.h"
+
+class TestMetaOutput : public TestRequiringOptions
+{
 protected:
-    std::string filename;
+    using TestRequiringOptions::TestRequiringOptions;
     tilegen::TilesetGenerator tg;
-    const OptionsManager optionsManager;
     const metaoutput::MetaOutput& mo;
     TestMetaOutput() :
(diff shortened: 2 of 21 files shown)


hryniuk:
After thinking about it for a while, `argv` is good enough for . It's tricky to make `boost::program_options` work with GTest arguments - I mean allowing different order of them.

WangscapeTest/TestMetaOutput.cpp
```diff
@@ -12,17 +12,18 @@
...
 #include <tilegen/partition/TilePartitionerSquares.h>
 
-class TestMetaOutput : public ::testing::Test {
+#include "TestRequiringOptions.h"
+
+class TestMetaOutput : public TestRequiringOptions
+{
 protected:
-    std::string filename;
+    using TestRequiringOptions::TestRequiringOptions;
```
hryniuk: This using clause and a succeding constructor call is more vague than it could be. That's true, there could be a fixture for tests using `Options`, but I think setting and reading an options filename could be taken out from this class. Note that it could be completely hidden:
```cpp
// OptionsFilename.h
std::string getOptionsFilename();
void setOptionsFilename(std::filename);

// OptionsFilename.cpp

namespace
{
std::string filename;
}

std::string getOptionsFilename()
{
    return filename;
}

void setOptionsFilename(std::string optionsFilename)
{
    filename = optionsFilename;
}
```

Then `setFilename(filename)` would be called in `main()` and `getFilename` would be used by `TestRequiringOptions`.

What do you think about it?

serin-delaunay: Yes, that's probably better. Although I would prefer to keep the check that prevents `filename` from being set more than once. GTest tests are after all supposed to be independent.

serin-delaunay: Yes, that's probably better. I would prefer to keep the guard which prevents setOptionsFilename from being successfully called more than once, though; GTest tests are supposed to be unable to affect each other.

serin-delaunay:
I think this is ready to merge.

WangscapeTest/OptionsFilename.cpp
```diff
@@ -0,0 +1,21 @@
+#include "OptionsFilename.h"
+#include <stdexcept>
+
+namespace
+{
+std::string filename;
+bool filenameSet = false;
```
hryniuk: That's a good place for `boost::optional`, but it can stay as it is for now.

WangscapeTest/OptionsFilename.cpp
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const std::string& getOptionsFilename()
+{
+    return filename;
+}
+
+void setOptionsFilename(const std::string & optionsFilename)
+{
+    if (filenameSet)
```
hryniuk: Can we have brackets here 😃?

WangscapeTest/OptionsFilename.cpp
hryniuk: That's a good place for `boost::optional`, but it can stay as it is for now.

WangscapeTest/TestReseedable.cpp
```diff
@@ -3,7 +3,8 @@
 #include <noise/MakeReseedable.h>
 #include <noise/module/ModuleFactories.h>
 
-class TestReseedable : public ::testing::Test {
+class TestReseedable : public ::testing::Test
```
hryniuk: I got an assertion failure during this test:
```shell
[ RUN      ] TestReseedable.TestReseedableClamp
WangscapeTest: /home/hryniuk/projekty/Wangscape/lib/libnoise/src/module/clamp.cpp:50: void noise::module::Clamp::SetBounds(double, double): Assertion `lowerBound < upperBound' failed.
```

serin-delaunay: This is a bug in libnoise. The documentation for `noise::module::Clamp` ([link](http://libnoise.sourceforge.net/docs/classnoise_1_1module_1_1Clamp.html)) clearly states "The lower bound must be less than or equal to the upper bound.", but the assertion checks `<` rather than `<=`.

This is present in the official source on sourceforge, so I'm not sure why this code was working on my machine. Wangscape will need to support this use case, so I'll replace any usage of `Clamp` modules with a pair of `Min` and `Max` modules.

WangscapeTest/OptionsFilename.cpp
serin-delaunay: I've put the whole function in an `if`/`else` block now.

serin-delaunay:
Argh, I didn't put the new files in CMakeLists. The comment/review display on github can be really confusing. I'll open a new PR with the changes.

PR #56, opened 2017-01-03
Reseedable::curve and Reseedable::terrace

### LINKED ISSUE #53: Implement Reseedable::curve and Reseedable::terrace
The `Reseedable` class has methods and operators which allow it to be modified and combined with other `Reseedable`s. Module parameters are passed in these methods rather than set after construction. Most of the non-generator libnoise `Module`s have a corresponding method. Two libnoise modifier modules do not have a corresponding method: [Terrace](http://libnoise.sourceforge.net/docs/classnoise_1_1module_1_1Terrace.html) and [Curve](http://libnoise.sourceforge.net/docs/classnoise_1_1module_1_1Curve.html).

Both of these modules have a variable-length parameter set consisting of a set of control points. Normally these are set one at a time in separate function calls. Since `Reseedable::curve` and `Reseedable::terrace` would need to take all the parameters in one call, the control points should probably be passed as a pair of iterators.

serin-delaunay:
Resolves #53.

### DIFF
--- a/Wangscape/noise/Reseedable.h
+++ b/Wangscape/noise/Reseedable.h
@@ -58,7 +58,9 @@ private:
 template<typename InputIt>
 inline Reseedable Reseedable::terrace(InputIt first, InputIt last, bool inverted)
 {
-    // TODO static_assert that Iterator is at least an input iterator
+    static_assert(std::is_same<std::iterator_traits<InputIt>::iterator_category::input_iterator_tag,
+                  std::input_iterator_tag>::value,
+                  "InputIt must have input_iterator_tag");
     static_assert(std::is_same<typename std::iterator_traits<InputIt>::value_type,
                                double>::value,
                   "InputIt must yield values of type double");
@@ -74,7 +76,9 @@ inline Reseedable Reseedable::terrace(InputIt first, InputIt last, bool inverted
 template<typename InputIt>
 inline Reseedable Reseedable::curve(InputIt first, InputIt last)
 {
-    // TODO static_assert that Iterator is at least an input iterator
+    static_assert(std::is_same<std::iterator_traits<InputIt>::iterator_category::input_iterator_tag,
+                               std::input_iterator_tag>::value,
+                  "InputIt must have input_iterator_tag");
     static_assert(std::is_same<typename std::iterator_traits<InputIt>::value_type,
                                std::pair<double, double>>::value,
                   "InputIt must yield values of type std::pair<double, double>");
--- a/Wangscape/noise/Reseedable.h
+++ b/Wangscape/noise/Reseedable.h
@@ -61,7 +61,7 @@ inline Reseedable Reseedable::terrace(InputIt first, InputIt last, bool inverted
     // TODO static_assert that Iterator is at least an input iterator
     static_assert(std::is_same<typename std::iterator_traits<InputIt>::value_type,
                                double>::value,
-                  "Iterator must yield values of type double");
+                  "InputIt must yield values of type double");
     auto terrace_p = std::make_shared<module::Terrace>();
     terrace_p->InvertTerraces(inverted);
     std::for_each(first, last, [&terrace_p](const auto control_point)
@@ -77,7 +77,7 @@ inline Reseedable Reseedable::curve(InputIt first, InputIt last)
     // TODO static_assert that Iterator is at least an input iterator
     static_assert(std::is_same<typename std::iterator_traits<InputIt>::value_type,
                                std::pair<double, double>>::value,
-                  "Iterator must yield values of type std::pair<double, double>");
+                  "InputIt must yield values of type std::pair<double, double>");
     auto curve_p = std::make_shared<module::Curve>();
     std::for_each(first, last, [&curve_p](const auto control_point)
     {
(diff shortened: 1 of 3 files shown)


Wangscape/noise/Reseedable.h
```diff
@@ -43,6 +44,46 @@ struct Reseedable final
     Reseedable scalePoint(double x_scale, double y_scale, double z_scale);
     Reseedable translatePoint(double x_displace, double y_displace, double z_displace);
     Reseedable turbulence(double frequency, double power, int roughness, int seed);
+    template<typename Iterator>
+    Reseedable terrace(Iterator begin, Iterator end, bool inverted = false);
+    Reseedable terrace(int controlPointCount, bool inverted = false);
```
hryniuk: All these methods can be `const`.

Wangscape/noise/Reseedable.h
```diff
@@ -43,6 +44,46 @@ struct Reseedable final
...
+    template<typename Iterator>
+    Reseedable curve(Iterator begin, Iterator end);
+private:
+    Reseedable finaliseTerrace(std::shared_ptr<module::Terrace>& terrace_ptr);
+    Reseedable finaliseCurve(std::shared_ptr<module::Curve>& curve_ptr);
 };
 
+template<typename Iterator>
+inline Reseedable Reseedable::terrace(Iterator it, Iterator end, bool inverted)
```
hryniuk: These arguments bother me. What if `it` (should be named `begin` as in declaration) or `end` is not valid iterator anymore? Should `terrace` go through the whole container? If yes, I think it would be better to pass container as argument.

On the other hand, that's how std algorithms are declared, so it's not so bad :wink:.

Wangscape/noise/Reseedable.h
```diff
@@ -43,6 +44,46 @@ struct Reseedable final
...
     Reseedable translatePoint(double x_displace, double y_displace, double z_displace);
     Reseedable turbulence(double frequency, double power, int roughness, int seed);
+    template<typename Iterator>
+    Reseedable terrace(Iterator begin, Iterator end, bool inverted = false);
+    Reseedable terrace(int controlPointCount, bool inverted = false);
+    template<typename Iterator>
+    Reseedable curve(Iterator begin, Iterator end);
+private:
+    Reseedable finaliseTerrace(std::shared_ptr<module::Terrace>& terrace_ptr);
+    Reseedable finaliseCurve(std::shared_ptr<module::Curve>& curve_ptr);
```
hryniuk: Why are you passing `std::shared_ptr` by a reference? Is there any advantage?

Wangscape/noise/Reseedable.h
```diff
@@ -43,6 +44,46 @@ struct Reseedable final
...
+    // TODO static_assert that Iterator is at least an input iterator
+    static_assert(std::is_same<typename std::iterator_traits<Iterator>::value_type,
+                               double>::value,
+                  "Iterator must yield values of type double");
+    auto terrace_p = std::make_shared<module::Terrace>();
+    terrace_p->InvertTerraces(inverted);
+    for (; it != end; ++it)
+    {
+        terrace_p->AddControlPoint(*it);
+    }
```
hryniuk: That's perfect place to use `std::for_each`:

```cpp
std::for_each(it, end, [&p=terrace_p](auto t){p->AddControlPoint(*t);});
```

Wangscape/noise/Reseedable.h
```diff
@@ -43,6 +44,46 @@ struct Reseedable final
...
+    // TODO static_assert that Iterator is at least an input iterator
+    static_assert(std::is_same<typename std::iterator_traits<Iterator>::value_type,
+                               std::pair<double, double>>::value,
+                  "Iterator must yield values of type std::pair<double, double>");
+    auto curve_p = std::make_shared<module::Curve>();
+    for (; it != end; ++it)
+    {
+        const std::pair<double, double> value = *it;
+        curve_p->AddControlPoint(value.first, value.second);
+    }
```
hryniuk: This too (could be captured with `&curve_p`):

```cpp
std::for_each(it, end, [&p=curve_p](auto t){p->AddControlPoint(t->first, t->second);});
```

Wangscape/noise/Reseedable.h
serin-delaunay: They can't be const because they return an object which can call a non-const method (`setSeed`) of the current instance.

Wangscape/noise/Reseedable.h
serin-delaunay: Yes, my initial thought was to pass the container by const reference. But user code may want to use only part of a container, or an iterator that doesn't point to a container. I think the `std` method is best.

I'll fix the disagreement in parameter names between declaration and implementation, though.

Wangscape/noise/Reseedable.h
serin-delaunay: Yes.

Wangscape/noise/Reseedable.h
serin-delaunay: Yes, although the specification of

serin-delaunay: And `std::for_each` accounts for that, so it would be `AddControlPoint(t.first, t.second)`.

Wangscape/noise/Reseedable.h
serin-delaunay: Passing it by reference skips the reference count increment/decrement/comparison. But this part of the code isn't performance-sensitive (and it might even be optimised out), so it's probably not worthwhile.

Wangscape/noise/Reseedable.h
serin-delaunay: STL algorithms use `first` and `last`, so I should probably conform to that naming scheme. Even though `last` isn't really the last iterator dereferenced.

Issue #46, opened 2016-12-29
Use names and typenames to clarify and validate coordinate arithmetic


serin-delaunay:
Currently there are at least 4 different 2D vector types:
* Pixel coordinates in a bitmap (input or output, tile or tileset)
* Tile index in a tileset
* Real coordinates for noise evaluation
* Tile resolution

Many variables of these types are simply labelled `x` and `y` (and not packed into a structure), and pixel coordinates are not distinguished from tile indices.

This could be fixed with just name changes, but I think it would also be worthwhile to use typechecking to enforce correct arithmetic and usage of these variables. There don't seem to be any errors of this type at the moment, but it would be a good idea to defend against future errors.

serin-delaunay:
Aside from internal types, the JSON input and output files use inconsistent formats for 2D vector values. Either they should all use the same format, or values in the same class (resolution, pixel coordinates, or tile index) should use the same format. (https://github.com/Wangscape/Wangscape/pull/111#discussion_r110268273)

serin-delaunay:
Now that we have Armadillo included, I think `arma::Col<T>::fixed<2>` would be the most suitable class for this. It doesn't have `x` and `y` members like `sf::Vector`, but it does have a more complete set of operators (including pointwise and scalar).

serin-delaunay:
This can be made a Hacktoberfest issue after #163 is merged.

PR #64, opened 2017-01-07
Tileset iteration

serin-delaunay:
This adds a class, CartesianPower, and its iterator, which are intended to simplify the loop in `tilesetGenerator::generateClique` and any other code which may need to iterate over each tile in a tileset.

It more or less corresponds to `itertools.product(*[X]*n)` in Python, but in CartesianPower the fastest-changing element is the earliest. This can be changed if necessary.

Note that this is branched from #55 because functioning tests are nice.

TODO:
- [x] Write `CartesianPower` and `CartesianPowerIterator` classes.
- [x] Write tests.
- [x] Refactor `tilesetGenerator::generateClique` to use `CartesianPower`.
- [x] Provide a way to build or retrieve a container of values or const value references from a CartesianPowerIterator.
- [x] Make updates to that container efficient; it needn't be rebuilt after every iteration.
- [x] Make assertions about iterator classes used.
- [x] Move tile position calculation into `CartesianPowerIterator`.
- [x] Move `tilesetGenerator::calculateTilesetResolution` into `CartesianPower`.
- [x] Fix crash in `TestMetaOutput`
* ~Template the container used by `CartesianPowerIterator`?~ Nope.
* ~Allow more manipulation of `CartesianPowerIterator::mComponents`?~ Nope.
- [x] Make use of `<<` in `ipow` optional based on a default template parameter.
- [x] Update `ipow` tests to use cases where `IBase != IExp`, and with both values of `UseBitshift` parameter.
- [x] Reduce the number of template instantiations of `ipow` tests.
- [x] Change `PowInt` to `IPow` in test names.
- [x] Remove `five` from `ipow` tests.
- [x] Move `ipow` into `ipow.h`, maybe even in a separate repo (as it now has a submodule dependency).
- [x] Move `CoordinatePacker` tests into a new `TestCoordinatePacker.cpp` file.
- [x] Resolve merge conflicts.


hryniuk:
I really appreciate the idea of "early pull requests". I do so in my work, I've started to do it with my previous pull request here and I see that you like it too (we could add it to coming `CONTRIBUTING.md`) 😄.

I belive that such openness and fast feedback will help new contributors taking the first steps in Wangscape.

serin-delaunay:
Since this is branched from #55, it's probably best to wait until #55 is merged before merging this. Some of the requested changes aren't on this branch.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+inline std::pair<size_t, size_t> CartesianPowerIterator<InputIt>::coordinates_2d() const
+{
+    size_t clique_size = std::distance(getFirst(), getLast());
+    size_t x = 0; size_t y = 0;
+    for (size_t i = 0; i < static_cast<size_t>(CORNERS); i++)
+    {
+        size_t& z = ((i + 1) % 2) ? x : y;
+        z *= clique_size;
+        z += coordinate(i);
+    }
```
hryniuk: Oh, we've discussed it already. Now I think the best could be splitting this `for` into two `for` loops, one for `x` and one for `y`. And maybe a function for these lines:
```cpp
z *= clique_size;
z += coordinate(i);
```
What do you think? How we can make it easier to understand?

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -30,7 +28,8 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
     boost::filesystem::path p(options.relativeOutputDirectory);
     for (const auto& clique : options.cliques)
     {
-        std::pair<size_t, size_t> tileset_resolution = calculateTilesetResolution(clique.size());
+        CornersGenerator cp(clique, static_cast<size_t>(CORNERS));
```
hryniuk: Find a move descriptive name for it, please. `cornerGenerator` is good enough. You don't want to wonder, what does `cp` mean e.g. next month :wink:.

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -51,42 +50,16 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
...
-            }
-            else
-            {
-                stop = false;
-                t = clique[i];
-                break;
-            }
-        }
+        const auto& corner_terrains = *it;
+        std::pair<size_t, size_t> tp = it.coordinates_2d();
```
hryniuk: What's `tp`? Could you rename it?

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -51,42 +50,16 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
...
+        TileGenerator::generate(image, tp.first, tp.second, corner_terrains,
+                                images, options, *mTilePartitioner.get());
+        mo.addTile(corner_terrains, filename,
```
hryniuk: Why adding a tile requires passing a filename? It's really strange for me now and I think we should have a second look on `MetaOutput`'s interface after merging this PR.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+        {
+            if (mIterators[i] == getLast())
+            {
+                mIterators[i] = getFirst();
+                mValues[i] = *mIterators[i];
+            }
+            else
+            {
+                mValues[i] = *mIterators[i];
+                break;
```
hryniuk: Why is this `break` here?

Wangscape/tilegen/CartesianPower.h
```diff
@@ -0,0 +1,106 @@
...
+
+template<typename InputIt>
+inline size_t CartesianPower<InputIt>::size() const
+{
+    size_t base_size = std::distance(first, last);
+    size_t result = 1;
+    for (size_t i = 0; i < power; i++)
+    {
+        result *= base_size;
+    }
```
serin-delaunay: `std::pow` doesn't accept integer arguments. Using it would mean converting to `double` and back to `size_t`.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+    std::vector<InputIt> mIterators;
+    std::vector<typename InputIt::value_type> mValues;
+};
+
+template<typename InputIt>
+inline CartesianPowerIterator<InputIt>::CartesianPowerIterator(InputIt first, InputIt last, InputIt init, size_t power) :
+    mFirst(first), mLast(last),
+    mIterators(power, init)
+{
+    if (init != last)
```
serin-delaunay: Yes. `init == last` is used when creating an `end` iterator.

Wangscape/tilegen/CartesianPower.h
```diff
@@ -0,0 +1,106 @@
...
+{
+public:
+    static_assert(std::is_base_of<std::input_iterator_tag, 
+                                  typename std::iterator_traits<InputIt>::iterator_category>::value,
+                  "InputIt must be at least an InputIterator");
+    typedef CartesianPowerIterator<InputIt> const_iterator;
+
+    CartesianPower(InputIt first_, InputIt last_, size_t power_);
+    template<typename Container>
+    CartesianPower(const Container& container, size_t power_);
```
serin-delaunay: `container` is not used to initialise a data member named `container`; is it your intention that all constructor arguments should have a trailing `_`?

Wangscape/tilegen/CartesianPower.h
```diff
@@ -0,0 +1,106 @@
...
+    const_iterator cbegin() const;
+    const_iterator cend() const;
+    const_iterator begin() const;
+    const_iterator end() const;
+
+    size_t size() const;
+    std::pair<size_t, size_t> size_2d(size_t resolution) const;
+
+    InputIt first;
+    InputIt last;
```
serin-delaunay: `first` and `last` can be any pair of iterators that define a valid range (that is, `last` should be reachable from `first`). They needn't be the `begin()` and `end()` of a container.

Changing `first`, `last`, or `power` will not invalidate any associated  `CartesianPowerIterator`s, but could allow `CartesianPower` instances to be reused or initialised outside the constructor. I don't think it does any harm to make them non-const, and future code may need to change them.

Wangscape/tilegen/CartesianPower.h
hryniuk: Yeah, I know, I thought about these conversions. If you don't want to use `std::pow`, maybe there should be a version for size_t in Wangscape. What do you think? Just to avoid duplication.

Wangscape/tilegen/CartesianPower.h
hryniuk: Yes, I thought about consistent naming. BTW shouldn't these members (`first`, `last`, `power` be private)?

Wangscape/tilegen/CartesianPower.h
serin-delaunay: A `pow` for integral types would be useful. Probably in `CartesianPowerIterator.h` or a new `utils.h` for these kinds of functions.

Wangscape/tilegen/CartesianPowerIterator.h
serin-delaunay: If we didn't have this line, we would count from `0000` to `1111`, etc. With this line, we count from `0000` to `1000`, etc.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+        for (const auto& it : mIterators)
+            mValues.push_back(*it);
```
serin-delaunay: An initial draft just stored the iterators, and had a function wrapping `std::transform` to output the vector of values. However, this meant needlessly copying all the values every time they were accessed. On the majority of iterations, only one value changes.

With a vector of copies of the values, we make on average one copy per iteration, and the client code can make the decision to copy the whole vector if necessary.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+{
+    if (getFirst() != other.getFirst())
+        return false; // different ranges
+    if (getLast() != other.getLast())
+        return false; // different ranges
+    if (getIterators().size() != other.getIterators().size())
+        return false; // different powers
+    if (*getIterators().crbegin() == getLast() &&
+        *other.getIterators().crbegin() == getLast())
+        return true; // both are cend
```
serin-delaunay: The "both are `cend`" comment is necessary because it's the only line of code that clearly defines what it means for a `CartesianPowerIterator` to be equal to `cend()`. If I move that check into a function `isEnd()` then all these comments can be removed.

Wangscape/tilegen/CartesianPower.h
serin-delaunay: OK, I guess "all constructor parameters have a trailing `_` and all member names do not" is a simpler style guideline.

`power` could be made private; a warning should be emitted when trying to set it to a value other than 3, 4, or 6.

It's not really possible to validate the iterators dynamically. If `size` or `base_size` were made a data member of `CartesianProduct`, the iterators should be made private to enforce updates to `size`. Currently that's not the case.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+}
+
+template<typename InputIt>
+inline bool CartesianPowerIterator<InputIt>::operator!=(const CartesianPowerIterator& other) const
+{
+    return !(*this == other);
+}
+
+template<typename InputIt>
+inline const typename CartesianPowerIterator<InputIt>::Values& CartesianPowerIterator<InputIt>::operator*() const
```
serin-delaunay: Altering the vector of stored values would not just affect the current iteration, but all future iterations until the altered values are overwritten. I don't think it would be a good idea to allow that. If the user needs to alter the output values, they can make a mutable copy.

Wangscape/tilegen/TilesetGenerator.cpp
serin-delaunay: The `tiles` field of `MetaOutput` describes the content of a tile (the four corner terrains) and its location (the tileset and offset within the tileset).

The tileset is identified by filename; the only alternatives are the filename with the extension removed, and the list of terrains in the tileset. I prefer the full filename, as it's more explicit.

Wangscape/tilegen/CartesianPowerIterator.h
serin-delaunay: Perhaps it should be a class:
```
CompoundCoordinate x(base_size);
CompoundCoordinate y(base_size);
for(int i=0; i < static_cast<size_t>(CORNERS); i += 2)
    x.addCoordinate(coordinate(i));
for(int i=1; i < static_cast<size_t>(CORNERS); i += 2)
    y.addCoordinate(coordinate(i));
return {x.coordinate(), y.coordinate()};
```

Wangscape/tilegen/TilesetGenerator.cpp
serin-delaunay: It's the position of the tile within the tileset - counting tiles, not pixels. I still haven;t decided on the best name for that kind of coordinate (see #46). I'll call it `tile_position` for now.

Wangscape/tilegen/CartesianPower.h
```diff
@@ -0,0 +1,106 @@
...
+{
+
+template<typename InputIt>
+class CartesianPower
+{
+public:
+    static_assert(std::is_base_of<std::input_iterator_tag, 
+                                  typename std::iterator_traits<InputIt>::iterator_category>::value,
+                  "InputIt must be at least an InputIterator");
+    typedef CartesianPowerIterator<InputIt> const_iterator;
```
serin-delaunay: `CartesianPowerIterator` is not intended to ever provide non-const access to the values referenced by underlying iterator range, and it's not necessary to provide non-const access to the  vector of copies stored in the `CartesianPowerIterator`. The current implementation reduces copying by reusing this vector between iterations, so providing non-const access to it is especially undesirable (https://github.com/Wangscape/Wangscape/pull/64#discussion_r96147811).

I think that considering the intention of the class, it's more appropriate to call it a `const_iterator` than an `iterator`. Note that `string_view` has `typedef const_iterator iterator` ([cppreference](http://en.cppreference.com/w/cpp/string/basic_string_view#Member_types))

I can change its API to make the constness more explicit (specify const `reference` and `pointer` types in `iterator_traits`, make `operator*` return a `reference`), and I can implement more of the `InputIterator` concept. It lacks:
* `operator->`
* `operator++(int)`
* `CopyAssignable`
* `CopyConstructible`

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+{
+    return mValues;
+}
+
+template<typename InputIt>
+inline const typename InputIt::value_type& CartesianPowerIterator<InputIt>::get(size_t n) const
+{
+    return mValues[n];
```
serin-delaunay: Or `getValues()[n]`.

Wangscape/tilegen/CartesianPower.h
```diff
@@ -0,0 +1,106 @@
...
+namespace tilegen
+{
+
+template<typename InputIt>
+class CartesianPower
+{
+public:
+    static_assert(std::is_base_of<std::input_iterator_tag, 
+                                  typename std::iterator_traits<InputIt>::iterator_category>::value,
+                  "InputIt must be at least an InputIterator");
```
serin-delaunay: The `first` iterator is used in code equivalent to the following:
```
x = first;
*x;
++x;
*x;
x = first; // *
*x;
```
If `first` is an `InputIterator`, `first` may be invalidated before line `*` is called.

The template parameter for `CartesianPower` and `CartesianPowerIterator` should be renamed `FowardIt` and `static_assert`s should check for `std::forward_iterator_tag`, which supports multi-pass iteration.

Wangscape/tilegen/CartesianPower.h
```diff
@@ -0,0 +1,106 @@
...
+template<typename InputIt>
+typename CartesianPower<InputIt>::const_iterator CartesianPower<InputIt>::end() const
+{
+    return cend();
+}
+
+template<typename InputIt>
+inline size_t CartesianPower<InputIt>::size() const
+{
+    size_t base_size = std::distance(first, last);
```
serin-delaunay: `last` is currently used in `CartesianPowerIterator` to detect components that need to be reset, and to compare with `CartesianPower::cend()` - not just to calculate sizes. We can remove `last` from `CartesianPower`, if `CartesianPowerIterator` stores a `std::vector<size_t>` containing the offset of each iterator. This is how the original loop in `TilesetGenerator::generateClique` worked, so I would be happy with that solution.

The alternatives would be:
1. Calculate `last` with `std::advance` and store it. I don't like this solution, because the methods which require `last` (`operator++`, `operator==`) are used repeatedly when iterating over a `CartesianPower`, whereas the methods which require `base_size` (`size`, `size_2d`, `coordinates_2d`) are all "optional" - they do not need to be used when iterating. For this reason `base_size` should be an optional value and `last` should be primary. (With a vector of offsets this would not be the case)

2. Calculate `base_size` with `std::distance` and store it. `base_size` is only used in "optional" methods, so I don't think it makes sense to automatically calculate it in the constructor. It could be stored in a `boost::optional`.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,161 @@
...
+    if (getLast() != other.getLast())
+        return false; // different ranges
+    if (getIterators().size() != other.getIterators().size())
+        return false; // different powers
+    if (*getIterators().crbegin() == getLast() &&
+        *other.getIterators().crbegin() == getLast())
+        return true; // both are cend
+    for (size_t i = 0; i < getIterators().size(); i++)
+        if (getIterator(i) != other.getIterator(i))
+            return false; // different phases
```
serin-delaunay: or just `std::vector::operator==`.

serin-delaunay:
Argh. I tried to rebase it on master to remove the commits from `test-options-filename` but instead got several commits from master and duplicated all the commits from this branch.

Wangscape/tilegen/CoordinatePacker.h
```diff
@@ -0,0 +1,50 @@
...
+#include <assert.h>
+
+namespace tilegen
+{
+
+template<typename I>
+class CoordinatePacker
+{
+public:
+    CoordinatePacker(I size_);
```
hryniuk: There should be at least `is_arithmetic` type_trait for `I` as you're comparing it with numbers and using `operator*=` and `operator+=` with `I`-typed members.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+
+template<typename I>
+I _pow(I base, I exp)
+{
+    I result = 1;
+    for (int i = 0; i < std::numeric_limits<I>::digits; i++)
+    {
+        if (exp & 1)
+            result *= base;
+        else if (exp == 0)
```
hryniuk: You should just use `while` here.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+    else if (exp > std::numeric_limits<I>::digits)
+        throw std::range_error("Integer pow() with exp > digits and |base| > 1");
+    else if (base == -2)
+        return ((I)1 << exp) * (exp % 2 ? -1 : 1);
+    else if (exp >= std::numeric_limits<I>::digits)
+        throw std::range_error("Integer pow() with exp >= digits and base not in {-2, -1, 0, 1}");
+    else if (base == 2)
+        return (I)1 << exp;
+    return _pow(base, exp);
+}
```
hryniuk: Wow, that's the most compilicated `pow` functions I've ever seen 😆.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+
+template <typename I>
+typename std::enable_if<std::is_unsigned<I>::value, I>::type pow(I base, I exp)
+{
+    if (exp == 0)
+        return 1;
+    else if (base == 0)
+        return 0;
+    else if (base == 1)
+        return 1;
```
hryniuk: I don't think these two `else if` conditions are necessary. It would be just `_pow(0, exp)` and `_pow(1, exp)`.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+#include <type_traits>
+#include <stdexcept>
+
+namespace
+{
+
+template<typename I>
+I _pow(I base, I exp)
```
hryniuk: I think, basing on integers ranks, that `_pow(long long, long long)` would be enough.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+    else if (base == -1)
+        return exp % 2 ? -1 : 1;
+    else if (exp < 0)
+        throw std::domain_error("Integer pow() with exp < 0 and |base| != 1");
+    else if (base == 0)
+        return 0;
+    else if (exp > std::numeric_limits<I>::digits)
+        throw std::range_error("Integer pow() with exp > digits and |base| > 1");
+    else if (base == -2)
+        return ((I)1 << exp) * (exp % 2 ? -1 : 1);
```
hryniuk: `else if (base == 0)`
`else if (base == 1)`
`else if (base == -1)`
`else if (base == -2)`
are redundant. In fact, if you write `pow(long long, long long)`, you wouldn't need these two `pow` functions.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,262 @@
...
+        count++;
+    EXPECT_EQ(81, count);
+}
+
+TEST_F(TestCartesianPower, TestCartesianPowerComparisonBeginEnd)
+{
+    EXPECT_EQ(cp.cbegin(), cp.cbegin());
+    EXPECT_NE(cp.cend(), cp.cbegin());
+    EXPECT_NE(cp.cbegin(), cp.cend());
+    EXPECT_EQ(cp.cend(), cp.cend());
```
hryniuk: What are you checking here? If `cend()`/`cbegin` will return the same value twice? It's not needed.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,262 @@
...
+    typedef std::vector<int> Container;
+    typedef std::vector<int>::const_iterator Iterator;
+
+    Container base{4, 5, 6};
+    const tilegen::CartesianPower<Iterator> cp{base, 4};
+    TestCartesianPower()
+    {
+
+    };
+    ~TestCartesianPower() {};
```
hryniuk: Why do you need to define a constructor and a destructor?

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    for (size_t i = 0; i < values.size(); i++)
+    {
+        EXPECT_EQ(values[i], *it.getIterator(i));
+    }
+
+    for (size_t i = 0; i < values.size(); i++)
+    {
+        EXPECT_EQ(values[i], *it.getIterators()[i]);
```
hryniuk: It's redundant, you're checking internal data type here.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    tilegen::CoordinatePacker<size_t> coord_packer2(2);
+    EXPECT_ANY_THROW(coord_packer0.addCoordinate(-1));
+    EXPECT_ANY_THROW(coord_packer0.addCoordinate(2));
+    EXPECT_ANY_THROW(coord_packer0.addCoordinate(3));
+
+    tilegen::CoordinatePacker<size_t> coord_packer10(10);
+    EXPECT_ANY_THROW(coord_packer0.addCoordinate(-1));
+    EXPECT_ANY_THROW(coord_packer0.addCoordinate(10));
+    EXPECT_ANY_THROW(coord_packer0.addCoordinate(11));
```
hryniuk: It seems that you're checking the same test case four times. Why?

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,262 @@
...
+TEST_F(TestCartesianPower, TestCoordinatePackerSize)
+{
+    tilegen::CoordinatePacker<size_t> coord_packer0(0);
+    EXPECT_EQ(0, coord_packer0.size());
+    tilegen::CoordinatePacker<size_t> coord_packer1(1);
+    EXPECT_EQ(1, coord_packer1.size());
+    tilegen::CoordinatePacker<size_t> coord_packer2(2);
+    EXPECT_EQ(2, coord_packer2.size());
+    tilegen::CoordinatePacker<size_t> coord_packer10(10);
+    EXPECT_EQ(10, coord_packer10.size());
```
hryniuk: The same here - you're just checking if returned size is correct, I don't think it needs to be checked four times.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,262 @@
...
+    EXPECT_EQ(size, tilegen::CartesianPower<Iterator>(base, 3).size_2d(5));
+}
+
+TEST_F(TestCartesianPower, TestCartesianPowerCoordinates2D)
+{
+    auto it = cp.cbegin();
+    std::advance(it, 15);
+    std::pair<size_t, size_t> coordinates{1, 6};
+    EXPECT_EQ(coordinates, it.coordinates_2d());
+}
```
hryniuk: Rethink these tests: if all of them are needed, what you'd like to check and if you're testing all edge cases.

WangscapeTest/TestPowInt.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+    EXPECT_ANY_THROW(pow(this->two, this->digits));
+    EXPECT_ANY_THROW(pow(this->two, this->max));
+    EXPECT_ANY_THROW(pow(this->three, this->digits));
+    EXPECT_ANY_THROW(pow(this->three, this->max));
+    EXPECT_ANY_THROW(pow(this->digits, this->digits));
+    EXPECT_ANY_THROW(pow(this->digits, this->max));
+    EXPECT_ANY_THROW(pow(this->max, this->digits));
+    EXPECT_ANY_THROW(pow(this->max, this->max));
+}
```
hryniuk: Same here. I don't think there's a real difference e.g. between these lines (in terms of what they're testing):
```cpp
EXPECT_EQ(this->one, pow(this->one, this->two));
EXPECT_EQ(this->one, pow(this->one, this->three));
EXPECT_EQ(this->one, pow(this->one, this->five));
EXPECT_EQ(this->one, pow(this->one, this->digits));
```

WangscapeTest/TestPowIntSigned.cpp
```diff
@@ -0,0 +1,102 @@
...
+    const static I minus_three = -3;
+    const static I minus_two = -2;
+    const static I minus_one = -1;
+    const static I zero = 0;
+    const static I one = 1;
+    const static I two = 2;
+    const static I three = 3;
+    const static I five = 5;
+    const static I digits = std::numeric_limits<I>::digits;
+    const static I max = std::numeric_limits<I>::max();
```
hryniuk: Oh, `const`, `static`... Is `static` needed? I would resign from these names. Checking limits, one negative number, one positive, `0`, `1`, `-1` should be enough.

WangscapeTest/TestPowIntSigned.cpp
```diff
@@ -0,0 +1,102 @@
...
+    TestPowIntSigned()
+    {
+
+    };
+    ~TestPowIntSigned() {};
```
hryniuk: Do you need these methods?

Wangscape/tilegen/CartesianPower.h
hryniuk: I don't know. With such guideline, it could become over-complicated: members should have `m` prefix, constructor parameters should have trailing `_`, other methods' arguments should have "normal" naming, but if the're setting one of members, one should then find a new name or use this ugly `this->member` notation.

The best guideline I've ever seen was adding trailing `_` to **all** members, both `private` and `public`. It could be `m` prefix as well. This way argument names will be consistent and they will differ from member names enough to notice, which variable is a member.

Wangscape/tilegen/CartesianPower.h
hryniuk: So maybe it would be better to make them private and change with methods?

Wangscape/tilegen/CartesianPowerIterator.h
hryniuk: Can they? It returns reference to `const`.

Wangscape/tilegen/CartesianPowerIterator.h
hryniuk: Seems quite reasonable.

Wangscape/tilegen/CoordinatePacker.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+#include <assert.h>
```
serin-delaunay: I removed the assertions I used in this file because failing an assertion outside of `(EXPECT|ASSERT)_DEATH` causes GTest to crash. This `#include` should just be removed. Also, in future code assertions should only be used for cases that cannot occur in normal use or in tests.

serin-delaunay: (But if I do add assertions anywhere, I'll use `<cassert>`)

Wangscape/noise/Reseedable.cpp
```diff
@@ -96,7 +92,19 @@ Reseedable Reseedable::operator+(double bias)
     return makeReseedable(result);
 }
 
-Reseedable Reseedable::operator-(Reseedable & other)
+Reseedable & Reseedable::operator+=(Reseedable & other)
```
serin-delaunay: This change comes from one of the commits accidentally introduced during rebase, and was already merged. Also, once again it can't be `const` because the returned `Reseedable` instance has a non-const method that can alter `other`. `other` should be passed by value, but that's fixed in #73.

Wangscape/tilegen/CoordinatePacker.h
serin-delaunay: Probably it should be `is_integral`, because the resulting packed coordinates are not unique when floating-point values are used.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,255 @@
...
+}
+
+template<typename ForwardIt>
+inline typename CartesianPowerIterator<ForwardIt>::reference CartesianPowerIterator<ForwardIt>::getValues() const
+{
+    return mValues;
+}
+
+template<typename ForwardIt>
+inline const typename ForwardIt::value_type& CartesianPowerIterator<ForwardIt>::getValue(size_t n) const
```
serin-delaunay: If we just have one, I would prefer to have `getValue`, as its meaning is clearer. `operator[]` only indicates its meaning from its return type.

Wangscape/utils.h
serin-delaunay: `_pow(x, std::numeric_limits<I>::max())` with `x=0` or `x=1` would execute a loop with `std::numeric_limits<I>::digits` iterations.

hryniuk: For `long long` that will be **only** 63 iterations, I don't think it's a big problem, cause it shouldn't happen too often. Anyway, `base == 0` and `base == 1` are not unsigned types specific, so, if I'm right,  moving these checks to `_pow` won't break anything and will make it shorter and consistent.

serin-delaunay: The order of conditions is important. `exp == 0`, `base == 1` and `base == -1` must be checked before `exp < 0`, which must be checked before `base == 0`, which must be checked before the overflow conditions.

I can move the common checks into a function, but it'll be a new function, not `_pow`.

Wangscape/utils.h
serin-delaunay: All four conditions (and `base == 2`) replace the O(log_2(exp)) loop with an O(1) operation.

Wangscape/utils.h
serin-delaunay: It's simpler than both of Boost's implementations (one for multiprecision numbers, one for `exp` known at compile time) and an example `double` implementation (http://www.netlib.org/fdlibm/e_pow.c). I tried to find CPython's implementation, but it's buried under a mountain of `#define`s and pointers. :P

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: They can both be removed.

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: Copypaste fail. This should be replaced with a loop that checks the correct object in each case.

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: Now that coordinate packing is moved into a tested class, this test can be replaced with one that just checks uniqueness of the set of generated coordinate pairs.

WangscapeTest/TestPowIntSigned.cpp
serin-delaunay: They can be removed.

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: They produce different function calls: `cend().operator==(cbegin())` and `cbegin().operator==(cend())`. `CartesianPowerIterator`s with `isEnd() == true` are quite different internally from dereferenceable `CartesianPowerIterator`s, so it's worthwhile to check that comparison operators work correctly no matter which instance is `*this`.

Wangscape/utils.h
serin-delaunay: I think the decision to convert to/from wider datatypes before using `pow` should be up to the user code. Also, there are types with more digits than `long long` (`unsigned long long`, multiprecision types).

Wangscape/utils.h
serin-delaunay: The `for` condition is meant to prevent the shifted sign bit from being used in `exp & 1`. But the wrapping function already ensures `exp > 0` here, so `while` can be used.

WangscapeTest/TestPowIntSigned.cpp
serin-delaunay: The `TYPED_TEST` macros access `I` as `TypeParam`. I would prefer not to have all the `IntPow` tests make calls like `EXPECT_EQ((TypeParam)1, pow((TypeParam)0, (TypeParam)0));`, nor have to put `typedef TypeParam I;` at the start of every test case. `this->zero` is not ideal, but I prefer it to the alternatives. 

I could try giving `TestPowInt<I>` a member method `I callPow(I base, I exp) {return pow<I>(base, exp);}`. Then the test cases could do something like `EXPECT_EQ((TypeParam)1, this->callPow(0, 0));`. That would still have one `TypeParam` per line, though.

I can reduce the number of checks made, but they're fast enough not to have any cost associated. `base == 2` and `base == -2` are treated differently in `pow`, so `3` and `-3` will still be needed in some tests. `5` can probably be removed. `digits` and `max` are limits, although some of the combinations they're used in can be removed.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+        exp >>= 1;
+        base *= base;
+    }
+    return result;
+}
+
+}
+
+template <typename I>
+typename std::enable_if<std::is_unsigned<I>::value, I>::type pow(I base, I exp)
```
serin-delaunay: `pow(float, float)` and `pow(double, double)` seem to be in the global namespace, not just `std`, and ambiguous calls like `int i; size_t s; pow(i, s);` cause the compiler to try to match the `std` version instead of this. This should be renamed to something like `ipow`.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+namespace
+{
+
+template<typename I>
+I _pow(I base, I exp)
+{
+    I result = 1;
+    for (int i = 0; i < std::numeric_limits<I>::digits; i++)
+    {
+        if (exp & 1)
```
serin-delaunay: This should be `if (exp % 2)`.

Wangscape/utils.h
```diff
@@ -0,0 +1,64 @@
...
+            result *= base;
+        else if (exp == 0)
+            break;
+        exp >>= 1;
```
serin-delaunay: This should be `exp /= 2;`.

Wangscape/utils.h
serin-delaunay: Ideally this should be `template <typename I, typename J> I ipow(I, J)`. GTest's `TYPED_TEST` only allows one type parameter, but this could be circumvented by using `std::pair<I, J>` as the test template parameter and unpacking it with `TypeParam::first_type`/`TypeParam::second_type`.

serin-delaunay: The function should have a comment documenting:
* Template argument requirements.
* Behaviour when invalid arguments are passed (in particular the behaviour for "non-obvious" overflows like `pow<char>(5, 5)`, which is undefined).

Wangscape/utils.h
serin-delaunay: Also, for the sake of clarity and compatibility with unexpected types it may be better to use named `constexpr` values instead of literal `0`, `1`, `2` in `_pow`.

WangscapeTest/TestPowInt.cpp
serin-delaunay: `three` and `five` can be removed.

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: 0, 1, and 2 are all corner cases, and 10 represents the generic case. I don't expect implementations of `CoordinatePacker::size()` to differ between these values, and the current implementation treats all cases the same, but tests are meant to catch unexpected cases :)

serin-delaunay:
Managed to undo the rebase. The commits from #55 seem to have disappeared from this PR without doing the rebase, although they still show up in Github Desktop locally.

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: I think as long as `getValue`, `getValues`, `getIterator`, and `getIterators` are in `CartesianPowerIterator`'s public API, there should be tests to ensure that they're consistent.

I think it's possible to make `getIterator[s]` `private`, though.

WangscapeTest/TestPowInt.cpp
serin-delaunay: `TestPowIntTooLarge` can probably be reduced to just `2**digits` and `3**digits`. But again, there is no significant cost associated with these extra checks.

Wangscape/tilegen/CartesianPowerIterator.h
serin-delaunay: With `getLast()` removed and all the other compared values temporary, `std::tie` is no longer applicable to this case.

Wangscape/tilegen/CartesianPowerIterator.h
serin-delaunay: If `ForwardIt::value_type` is `CopyConstructible`, then the user should be able to make a copy of the value returned from `operator*`. `CartesianProductIterator` is only designed to be an `InputIterator`, which does not require the pointed values to be mutable (`std::iterator_traits<It>::reference` can be `const`).

What kind of use case do you have in mind for altering the output value vector?

Wangscape/tilegen/CartesianPower.h
serin-delaunay: `first`, `base_size` (which replaces `last`), and `power` are all independent variables which can be changed without any requirements or unexpected side-effects. I don't see what problem would be fixed by putting them behind `get`/`set` methods.

(`CartesianPower::size()`/`size_2d()` are much simpler operations now that the `std::distance` call is removed, and they are only expected to be called once for each `CartesianPower`, so I don't think it would be worthwhile to cache those values).

Wangscape/tilegen/CartesianPower.h
serin-delaunay: > ... other methods' arguments should have "normal" naming, but if the're setting one of members, one should then find a new name or use this ugly `this->member` notation.

I don't think this is really a problem. Almost all functions which set a data member will set a `protected` or `private` data member, which will have an `m` prefix. In the rare case when a member method sets a `public` data member, `this->member = member` will not detract from readability.

When designing a class API, my priorities for readability and consistency are:
1. What the user will have to type to use the class: class name, `public` method names, `public` data members, (`public` dependent types).
2. What the user will have to read to use the class: method parameter names, template parameter names, `public` dependent types.
3. What the user will have read and write to make derived classes: `public` and `protected` members.
4. Private implementation details.

Since method parameter names rank below data member names, it's the method parameter names that will have to gain trailing underscores when necessary. I would be happy to require a trailing underscore in all method parameter names, but at the moment (AFAIK) it's not necessary because we have no class methods used to set a public data member to the value of a parameter.

Wangscape/utils.h
hryniuk: For me type promotions are enough and a bit wider range of `unsigned long long` positive values irrelevant, but if you think it's important, keep this template :).

Wangscape/utils.h
hryniuk: Ok, but I would assume that `exp` should be positive (as result cannot be non-integer) and either make it `unsigned` or put assert with preconditions. You could consider it.

A new functions is a good option.

WangscapeTest/TestCartesianPower.cpp
hryniuk: I mean `EXPECT_EQ(cp.cend(), cp.cend());`. Can it fail?

Wangscape/utils.h
hryniuk: Ok, ok :D.

Wangscape/utils.h
hryniuk: I see, but:
1. It's not a big deal, it's some c ≤ 64 (128 maybe). I suppose that `exp` in Wangscape will be even ≤ 2^16.
2. It should be optimized by compiler.

I'd just like to let you know I think it's "premature optimization" and it makes this code vague in my opionion.

WangscapeTest/TestPowInt.cpp
hryniuk: Yes it is - you will have more code to maintain, which, in addition, is no really valuable, if these test cases are already covered in other lines. Just think of it :). There are a few bugs already and refactoring in progress. You'll (or we'll) have to spend a lot of time to fix it.

WangscapeTest/TestPowIntSigned.cpp
hryniuk: Ok, clear :).

Wangscape/utils.h
serin-delaunay: The current usage of `pow` in Wangscape is not in a performance-sensitive area (it's called twice per tileset), but `pow` for integral types is quite a common operation and I expect to reuse this function in future projects. When that happens, I would consider it less of an optimisation issue and more of a bug if I found that obvious O(1) cases were handled using the loop.

Detecting obvious results also allows them to be filtered out when detecting obvious exceptions (`exp >= digits` shouldn't throw when `abs(base) <= 1`, `exp < 0` shouldn't throw when `abs(base) == 1`).

I can imagine a very smart compiler detecting the redundant multiplications when `base` is `-1`, `0`, or `1`, but I wouldn't rely on that behaviour (and I doubt any current compiler would detect when an exponentiation-by-squaring algorithm can be replaced with a bitshift).

Wangscape/utils.h
serin-delaunay: The main reason to use `/` and `%` here is compatibility with custom types that might not have `>>`. In that case, it would probably best to add a `bool UseBitshift=true` to the template arguments so that the special cases for `base == 2` and `base == -2` can be turned off.

Wangscape/utils.h
serin-delaunay: Casting to `unsigned` or not should be the choice of the user code, not the `pow` function. Negative `exp` is valid when `base` is +-1. GTest interacts badly with `assert` and nothing stops user code from calling `pow` with negative arguments, so `throw` should be used instead. That's what the code already does.

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: `CartesianPowerIterator`s with `isEnd() == true` use different equality comparison conditions from those with `isEnd() == false`, so these test different lines of code.

Wangscape/utils.h
hryniuk: Hmm... Ok, you convinced me :).

WangscapeTest/TestCartesianPower.cpp
hryniuk: It's still too vague for me.There is the same object twice here, `cp`, and `cend()` is called two times, creating two iterators from the same set of arguments. They're compared with `operator==`, which internally check equality of them basing on each object's atrributes. They should have the same values (as both iterators were created from same arguments). Why should we check if `cp.cend() == cp.cend()`?

serin-delaunay: `cbegin() == cbegin()` is checked using criteria `a` and code path `a`. `cend() == cend()` is checked using criteria `b` and code path `b`. If criteria `b` is incorrect or code path `b` is incorrectly implemented, `cbegin() == cbegin()` might not reveal that mistake.

WangscapeTest/TestPowInt.cpp
serin-delaunay: `digits**digits` and `digits**max` can be removed.

While rewriting the conditionals used in `pow`, I broke `EXPECT_ANY_THROW` for `max**digits` and `max**max`, but not when `base` was 2, 3, or `digits`. A pretty clear demonstration of the value of those two tests.

With tests like [TestLinearMovingScaleBiasShort](https://github.com/Wangscape/Wangscape/blob/master/WangscapeTest/TestMovingScaleBias.cpp#L57) I would agree with you that extra tests bring too much maintenance work; each test assertion requires some visualisation and mental arithmetic to verify. However, with test targets as simple as `pow` and test assertions as simple as "an error should be thrown when `base >= 2` and `exp >= digits`", I don't think that's really a valid concern.

Wangscape/utils.h
serin-delaunay: Regarding extracting common checks into functions: the sequence of conditions where the checks must be made is:
```
(exp == 0) // all
else if (base == 1) // all
else if (base == -1) // base signed
else if (exp < 0) // exp signed
else if (base == 0) // all
else if (exp > std::numeric_limits<I>::digits) // base signed
else if (base == -2) // base signed
else if (exp >= std::numeric_limits<I>::digits) // all
else if (base == 2) // all
```
With the current template setup (`I base, I exp`), separating out all of the common checks would mean splitting the function into 5 pieces. With `IBase base, IExp exp`, it would require 6. I tried to do this pairs of `enable_if` functions and the result was rather long and not very readable. If I use the current method with `IBase base, IExp exp`, it will require 2 more specialisations with a lot of duplicated code.

I've made a version with [static_if](https://github.com/wichtounet/cpp_utils/blob/master/static_if.hpp) (included as a submodule), which keeps everything in one function and is much more readable.

WangscapeTest/TestPowIntSigned.cpp
serin-delaunay: (and the `static` can be removed)

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: I also removed `getCoordinate(size_t)`, since it's just `getCoordinates()[size_t]`.

serin-delaunay:
Nearly there. Updated the checklist at the top of the page.

serin-delaunay:
I think that's everything but the merge done. `ipow` and its tests have changed quite a lot, though.

Wangscape/tilegen/CartesianPowerIterator.h
```diff
@@ -0,0 +1,247 @@
...
+public:
+    static_assert(std::is_base_of<std::forward_iterator_tag,
+                                  typename std::iterator_traits<ForwardIt>::iterator_category>::value,
+                  "ForwardIt must be at least a ForwardIterator");
+
+    typedef std::vector<typename ForwardIt::value_type> Values;
+    typedef std::vector<ForwardIt> Iterators;
+    typedef std::vector<size_t> Coordinates;
+    typedef const Values& reference;
+    typedef const Values* pointer;
```
hryniuk: I think `pointer` should be renamed to e.g. `ValuesPtr` (beginning with capital to stick to "convention" above). The same for `reference`.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+    EXPECT_EQ(81, cp.size()) <<
+        "CartesianProduct has incorrect size";
+}
+
+TEST_F(TestCartesianPower, TestCartesianPowerCount)
+{
+    size_t count = 0;
+    for (const auto& it : cp)
+        count++;
```
hryniuk: It looks like `std::accumulate`.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    EXPECT_EQ(cp.cbegin(), cp.begin());
+    EXPECT_EQ(cp.cend(), cp.end());
+}
+
+TEST_F(TestCartesianPower, TestCartesianPowerComparisonLast)
+{
+    auto it = cp.cbegin();
+    for (size_t i = 0; i < cp.size(); i++)
+        ++it;
```
hryniuk: Could you use `std::advance` here? It will be nice, if it works the same way as standard iterators.

WangscapeTest/TestCartesianPower.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+    static_assert(std::is_copy_constructible<tilegen::CartesianPowerIterator<Iterator>>::value,
+                  "CartesianPowerIterator not CopyConstructible");
+    static_assert(std::is_copy_assignable<tilegen::CartesianPowerIterator<Iterator>>::value,
+                  "CartesianPowerIterator not CopyAssignable");
+    auto it = base.cbegin();
+    // try to copy-construct
+    auto it2 = it;
+    // try to copy-assign
+    it = base.cend();
```
hryniuk: Is this test needed? It hasn't any runtime assert. Could these `static_assert`s be placed "around" `CartesianPowerIterator` class? Or it seems that there could be `EXPECT_EQ`, checking if assignment was succesful.

WangscapeTest/TestCoordinatePacker.cpp
```diff
@@ -0,0 +1,71 @@
+#include <gtest/gtest.h>
+
+#include <vector>
+#include <algorithm>
+
+#include <tilegen/CoordinatePacker.h>
+
+class TestCoordinatePacker : public ::testing::Test
+{
+protected:
```
hryniuk: It should be removed, I belive.

WangscapeTest/TestCoordinatePacker.cpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+        tilegen::CoordinatePacker<size_t> coord_packer(i);
+        EXPECT_ANY_THROW(coord_packer.addCoordinate(-1));
+        EXPECT_ANY_THROW(coord_packer.addCoordinate(i));
+        EXPECT_ANY_THROW(coord_packer.addCoordinate(i + 1));
+    }
+}
+TEST_F(TestCoordinatePacker, TestCoordinatePackerSize)
+{
+    for (auto i : {0, 1, 2, 10})
```
hryniuk: These values could be placed in some `TestCoordinatePacker` member.

WangscapeTest/TestCoordinatePacker.cpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+        tilegen::CoordinatePacker<size_t> coord_packer(i);
+        EXPECT_EQ(i, coord_packer.size());
+    }
+}
+
+TEST_F(TestCoordinatePacker, TestCoordinatePacker1)
```
hryniuk: Name `TestCoordinatePacker1` doesn't tell very much about this testcase. Could you find a better name?

hryniuk: The same for test cases below.

Wangscape/tilegen/CartesianPowerIterator.h
serin-delaunay: `reference` and `pointer` are [required](http://en.cppreference.com/w/cpp/concept/InputIterator) by the `InputIterator` concept. I could make `ValuesRef` and `ValuesPtr`, but they would only be used in `typedef ValuesRef reference` and `typedef ValuesPtr pointer`

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: I'm not sure that `size_t count = std::accumulate(cp.cbegin(), cp.cend(), 0, [](size_t acc) {return acc + 1;});` would be simpler here?

WangscapeTest/TestCartesianPower.cpp
serin-delaunay: These lines of code were meant to fail compilation if `CartesianPowerIterator` wasn't copy-assignable/constructible. Unfortunately they compiled and ran even when the `static_assert`s failed, so they can be removed.

I'll move the `static_assert`s into the `CartesianPowerIterator` class if possible.

WangscapeTest/TestCartesianPower.cpp
hryniuk: And `std::distance`?

Wangscape/tilegen/CartesianPowerIterator.h
hryniuk: Ok, didn't know.

serin-delaunay:
I've resolved the merge conflicts, but Wangscape is crashing when compiled with optimisations (not in debug mode). I'll have to fix this before pushing the changes.

Issue #51, opened 2016-12-31
Alpha underflow in linear alpha calculator?


serin-delaunay:
This seems to be a separate bug from #38. The four tiles in the left column of this tileset from `example3` all have four or five black pixels in a line. The other two tilesets generated on this run are all fine. This seems to be a rare bug, so it's worth preserving this first observation for later, when it can be properly investigated.

![s v](https://cloud.githubusercontent.com/assets/15715657/21579566/e3d7857a-cfaf-11e6-9358-a0e5603a28be.png)

serin-delaunay:
This is not a rare bug - I've seen a few more examples.

serin-delaunay:
It's still possible to write configurations that produce this issue, but considering that

* the linear alpha calculator now has a simpler alternative (top two) which can produce very similar visuals without black patches, and
* all module groups are fully customisable, and noise values can be clamped to be greater than zero before applying alpha calculation,

Any appearances of this bug can be considered faults in the configuration files rather than faults in Wangscape. Therefore this issue can be closed.

PR #68, opened 2017-01-09
No double-to-float conversions

serin-delaunay:
Corner weights from `TilePartitionerGradient` and `TilePartitionerPerlin` were being converted from doubles to floats before `AlphaCalculator` methods were applied. I don't think there's any good reason to do that. This PR removes all of those conversions.

These conversions were a suspected cause of #51. The issue remains unresolved, but the size and frequency of black patches may have been reduced.

### DIFF
--- a/Wangscape/tilegen/partition/TilePartitionerGradient.cpp
+++ b/Wangscape/tilegen/partition/TilePartitionerGradient.cpp
@@ -25,7 +25,7 @@ void TilePartitionerGradient::makePartition(TilePartition & regions,
                    mOptions.tileFormat.resolution);
         masks.push_back(img);
     }
-    std::vector<float> weights(corners.size(), 0.f);
+    std::vector<double> weights(corners.size(), 0.);
     const int resolution_sub_1 = mOptions.tileFormat.resolution - 1;
     const int quarter_res = mOptions.tileFormat.resolution / 4;
     alpha::AlphaCalculatorLinear ac;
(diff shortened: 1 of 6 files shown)


Wangscape/tilegen/partition/TilePartitionerGradient.cpp
```diff
@@ -25,18 +25,18 @@ void TilePartitionerGradient::makePartition(TilePartition & regions,
...
         for (size_t y = 0; y < mOptions.tileFormat.resolution; y++)
         {
-            weights[0] = (float)gradientWeight(x, y, 0, 0, quarter_res);
-            weights[1] = (float)gradientWeight(x, y, 0, resolution_sub_1, quarter_res);
-            weights[2] = (float)gradientWeight(x, y, resolution_sub_1, 0, quarter_res);
-            weights[3] = (float)gradientWeight(x, y, resolution_sub_1, resolution_sub_1, quarter_res);
+            weights[0] = (double)gradientWeight(x, y, 0, 0, quarter_res);
+            weights[1] = (double)gradientWeight(x, y, 0, resolution_sub_1, quarter_res);
+            weights[2] = (double)gradientWeight(x, y, resolution_sub_1, 0, quarter_res);
+            weights[3] = (double)gradientWeight(x, y, resolution_sub_1, resolution_sub_1, quarter_res);
```
hryniuk: I'm not sure if it must be casted, but if it does, use `static_cast<double>` here, please. One advantage is that it's easy to grep for, and the other is that it is safer (e.g. another version of "raw casting", `double(...)` can be treated by a compiler as a function declaration).

serin-delaunay: Done.

Wangscape/tilegen/partition/TilePartitionerPerlin.cpp
```diff
@@ -29,16 +29,16 @@ Reseedable TilePartitionerPerlin::makeCornerModule(const Corners& corners,
...
 
-void TilePartitionerPerlin::noiseToAlpha(std::vector<noise::RasterValues<float>>& noise_values,
+void TilePartitionerPerlin::noiseToAlpha(std::vector<noise::RasterValues<double>>& noise_values,
                                          std::vector<sf::Image>& outputs,
                                          size_t resolution) const
 {
-    std::vector<float> weights((int)CORNERS);
-    alpha::AlphaCalculatorMax ac;
+    std::vector<double> weights((int)CORNERS);
+    alpha::AlphaCalculatorLinear ac;
```
hryniuk: Why did you change `AlphaCalculatorMax` to `AlphaCalculatorLinear`?

serin-delaunay: #51 Only manifests when `AlphaCalculatorLinear` is used. I changed to `AlphaCalculatorLinear` in order to observe the effects of the changes in that use case.

In future the choice of `AlphaCalculator` should be in the options file (#37); at the current stage of the project I don't think it's very important which one is hard-coded.

PR #62, opened 2017-01-06
Tests, tilegen and reseedable CMakeLists

hryniuk:
Fixes first point from #59.

Issue #20, opened 2016-12-11
Fix autobuilds on appveyor


serin-delaunay:
~Building currently fails because rapidjson is not in the project's NuGet packages. The version of rapidjson currently in the NuGet gallery ([1.0.2](https://www.nuget.org/packages/rapidjson/)) is out of date, so I'm using a locally installed up-to-date version.~

~To fix this, either include rapidjson in the repository itself (maybe using a git submodule?), upload a new rapidjson version, or ask the owners of the current rapidjson package to update it (no response yet).~

The obstacles in this issue are rather different now (no rapidjson, no outdated NuGet packages), but the issue remains; Wangscape needs to be built and packaged in a Windows-based CI environment.

- [x] Build Wangscape and WangscapeTest.
- [x] Run tests.
* ~Upload test results.~ The [instructions](https://www.appveyor.com/docs/running-tests/#uploading-xml-test-results) given to do this have no effect.
* ~Pass `TestTilePartitionerSquares` or remove `TilePartitionerSquares` in favour of #108.~ Appveyor's OpenGL is too outdated to run Wangscape correctly (see #117). Skip `TestTilePartitionerSquares`.
- [x] Pass all other tests.
- [x] Add appveyor badge.
- [ ] Build standalone executables for Wangview and Wangcheck (with py2exe, PyInstaller, or cx_freeze): https://github.com/Wangscape/Wangcheck/issues/4, https://github.com/Wangscape/Wangview/issues/17.
- [ ] Produce standalone distributable package containing Wangscape[Test], Wangview, Wangcheck, and all necessary documentation/data files.
- [x] Run Wangscape.exe on example2 or example3, zip the output, and make it an artefact.

Caching built submodules will be a different issue.

PR #70, opened 2017-01-09
Add libnoise as submodule

hryniuk:
It's required to finish #62 and finally complete Travis configuration.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = lib/spotify-json
 	url = https://github.com/Wangscape/spotify-json.git
 	ignore = dirty
+[submodule "lib/libnoise"]
+	path = lib/libnoise
+	url = https://github.com/qknight/libnoise.git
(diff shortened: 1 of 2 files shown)


serin-delaunay:
I think the submodule should be in the folder `lib/libnoise/`.

hryniuk:
Yeah, I thought so too, but... **it will require changing all include directives and will have an impact on other PRs**. I'll do it, just a warning.

serin-delaunay:
Why would it affect include directives? If `lib/libnoise/src/` is added as an include folder, the current `#include <noise/noise.h>` directives should work fine.

hryniuk:
You're right, I don't know why I didn't notice it :frowning:.

hryniuk:
Ok, now it should be alright. I've also changed `CMakeLists.txt` in #62 to include `libnoise` instead of `noise`.

serin-delaunay:
Great :) I'll work on a PR to get the VS build using libnoise from the submodule.

hryniuk:
Wow, nice :). In an ideal world compilation would be the same on all platforms. I'll check cmake builds with CLion on Windows in my spare time.

serin-delaunay:
Once CI builds with CMake are working on https://appveyor.com or https://visualstudio.com (#20), I think it would be best to remove the VS stuff from the repository (I'll keep it for personal use). I'll start working on that once Travis builds are working. I've re-enabled Travis for Wangscape/Wangscape (transferring to the organisation had disabled it).

PR #58, opened 2017-01-05
Make module group

serin-delaunay:
Adds some convenience functions for building `ModuleGroup`s. Resolves one point from #57.

PR #73, opened 2017-01-11
Refactor Reseedable

### LINKED ISSUE #57: Refactor Reseedable
Currently `Reseedable` is a struct containing a `shared_ptr` to a `Module` and a `std::function` which (in theory) either does nothing or calls the `Module`'s `SetSeed` method.

The default factories for `Reseedable`s assign the correct `SetSeed` methods, but they do it by manual template specialisation of each individual `Module` subclass with a `SetSeed` method. Also, it would be trivial to invalidate the object by changing `setSeed` to a different function.

`Reseedable` should be replaced with a set of classes that:
* Has a template parameter that can be instantiated with any `Module` subclass.
* Contains an instance of or inherits from the template parameter.
* Has a common base class which:
  * Has a virtual `setSeed` method which ensures the correct behaviour.
  * Provides base reference access to the associated `Module` instance.

TODO moved into #73.

serin-delaunay:
Resolves #57. Not 64!

Changes:
* Replaces `ModulePtr` and `Reseedable` in namespace `noise` with `ReseedableBase`, `Reseedable<M>`, and `ReseedablePtr` in `noise::module`.
* `Reseedable<M>` chooses `setSeed` based on template parameter `M`.
* Moves all module-building operators and methods from `Reseedable` and `ModuleFactories.h/.cpp` into `ReseedableOps.h/.cpp`.

TODO:
- [x] Update tests to use new Reseedable class.
- [x] Rename tests that apply to function calls with altered syntax.
- [x] Fix `Curve` test ([reported here](https://gitter.im/Wangscape/Lobby?at=5876a07a873d96e16d5738cf)).
- [x] ~Forward module constructor arguments.~ Format `Module` subclasses according to libnoise style conventions; add default parameters, use only parameterless constructors.
- [x] Factor out code that builds `ModuleGroup` instances, using something like the changes in #58.
- [x] Give better names to some of the functions moved from `ModuleFactories`.
* Use an `enum` to distinguish different `ReseedableBase` subclasses and allow `ReseedablePtr`s to be correctly converted to `Reseedable<M>`? Without a compelling use case, this is probably a matter for a different PR.
* Choose the correct implementation of `SetSeed` (`T::SetSeed` or an empty function) without making an individual template specialisation for each `Module` subclass that has a `SetSeed` method. This should probably be a separate PR.
- [ ] Resolve merge conflicts.
- [x] Replace usages of `noise::module::Clamp` with pairs of `Min` and `Max` modules to avoid failing an incorrect assertion in `Clamp::SetBounds`. This is really one of the merge conflicts.
- [ ] Update CMakeLists.txt - this will have to be done while merging because this branch doesn't have `Wangscape/noise/[module/]CMakeLists.txt`.


serin-delaunay:
Regarding forwarding `Reseedable<M>` constructor arguments to `M`:

Some of the custom `Module` subclasses defined in Wangscape have constructor arguments, which with the current version of `Reseedable<M>` cannot be used in a call to `std::make_shared<Reseedable<M>>()`.  This could be fixed by using a template-based forwarding constructor.

On the other hand, Libnoise seems to object to constructor arguments, even optional ones, in `Module` subclasses. Most of the module-creating code is in functions that now reside in `ReseedableOps.cpp`, where a little extra verbosity is not harmful.

All considered, I think it would be best to remove the constructors with parameters, set default values for all the custom modules' parameters, and use libnoise-style conventions for `Module` subclasses. For instance, `NormLPQ::SetPQ` instead of `NormLPQ::setPQ`. The `Reseedable` API can use more concise methods and Wangscape style conventions.

serin-delaunay:
I've almost (!) figured out a way to deserialise libnoise module networks with spotify-json (#28), and while it's derived from this branch, probably most of the `Reseedable` API, including many of the lines changed in this PR, will be replaced or removed.
(`Reseedable` will move to `ModuleWrapper`, `ModuleGroup` may lose its `Module` base class, JSON deserialisation will supersede algebraic operators and functions on `ReseedablePtr`s)
So, I'm closing this PR.

Issue #80, opened 2017-01-26
Refactor/remove ModuleGroup


serin-delaunay:
Noise modules from libnoise may have source modules, which are specified using raw pointers. Together these form an implicit graph structure, which should be a directed acyclic graph (though this is not verified).

Wangscape currently hides this interface behind `ModuleGroup` modules, which contain a number of `Module`s accessible by name. `ModuleGroup::GetValue` looks up the output module and forwards the `GetValue` call. In theory the other modules in the group should be the output module's source modules (but this is not checked).

This system serves several purposes:
* Avoid interaction with raw pointers.
* Store complex module networks without allocating a container to store them all.
* Enable user code to call `SetSeed` on semantically related modules in one step using depth-first traversal.
* Enable user code to add modifying and combining modules to the network using simple arithmetic operators and functions.

However, it has some problems:
* There is no enforced connection between the modules stored in a `ModuleGroup` and the output module's sources.
* There is no convenient way to enforce a DAG structure on the module network.
* `ModuleGroup`s are not really `Module`s.
* The resulting module networks have a `ModuleGroup` between each pair of real `Module`s. That means that in noise evaluation (which is where most of Wangscape's time is spent), each real `Module::GetValue` call is interleaved with an unnecessary virtual function call and `map<string, _>` search.
* Many modules are only accessible by casting and dereferencing many layers of base class pointers, each of which could be a `ModuleGroup` or a normal module.

Deserialisation using spotify-json (#28) also seems to be a thorny issue. It doesn't look like the library can operate on raw pointers, so there are two options for making the module network usable:
1. Deserialise into a form from which the correct pointer structure can be deduced, and set the pointers afterwards. The current network structure, with decentralised ownership and difficulty in accessing deeply-nested modules, would not make this easy.
2. In the codec, use value-setting functions which incidentally make the appropriate `Set(Source|Control|[XYZ]Displace)Module` function calls. I would prefer this.

Ideally the JSON format used for noise module networks should be as compatible as possible with [RESTnoise](https://restnoise.readme.io/), which uses a flat network structure and identifies source modules by name. I think the only way to deserialise such a JSON format into a module network based on `ModuleGroup`s would be to have an intermediate datatype resembling a `rapidjson::document`.

To resolve this mess, I think the best course of action would be:
- [x] Remove `ModuleGroup`.
- [ ] Store modules in an explicit directed graph data structure.
- [ ] Store source module type (control, displace x/y/z, or source and index) as an edge property.
[...]

Issue #79, opened 2017-01-14
Move JSON schemas to doc/docs directory


hryniuk:

serin-delaunay:
Closed in #83.

PR #83, opened 2017-02-03
Decode Module Groups from JSON

### LINKED ISSUE #28: Build ModuleGroups from specs in options file

serin-delaunay:
Resolves #28, #57, #79; parts of #80, #21.

Changes:
* `Wrapper*` class added, replacing `Reseedable*` (but based on the new `Reseedable*` classes in #73).
* Codecs for `Wrapper`s.
* `ModuleGroup` modules replaced with a non-`Module` class which handles connecting input modules, quadrant selection, and reseeding. Cannot be encoded yet.
* Options file stores filenames of JSON-encoded module groups.
* Module group files and filenames added to examples.
* Tile generation uses module groups decoded from JSON.
* Tile generation reuses the combiner module group rather than rebuilding it 4 times per tile.
* Module groups decoded in multiple steps via `encoded_value_ref`s, for the following reasons:
  * `spotify::json::codec::one_of_t` with `cast_t` is the natural choice to decode polymorphic objects, but spotify-json documentation is not complete enough to make it clear how to do that in a way that's compatible with the `default_codec` API.
  * The resulting `one_of_t` would be a variadic template class with more than 28 arguments, and would not be dynamically extensible.
  * Direct conversion from JSON to correctly connected module groups is not practical in any case because module names may be referenced before the associated modules are decoded.
* Test decoding of all module types, and of a full module group.
* Bool field added to the options file to control whether debug output should be generated for noise modules. Debug output is overwritten for each corner of each tile. It goes in a "test" folder in the current working directory, and fails if the folder is not present. This can be changed if necessary, but a full implementation of #39 is beyond the scope of this PR.
* Examples and schemas moved to `doc` directory.
* Building has become rather slow.

To do:
- [x] Add codecs for all remaining module types:
  - [x] Standard libnoise modules
  - [x] Modules/pseudomodules defined by Wangscape:
    - [x] `CornerCombinerBase`
    - [x] `Exp`
    - [x] `Forward` (see below)
    - [x] `NormLPQ`
    - [x] `Pow`
    - [x] `QuadrantSelector`
    * ~`CornerCombiner`?, `EdgeFavouringMask`?~ I implemented the main module group in 25 modules using the module classes and codecs already defined; these two are probably not necessary.
    * ~`VariableScaleBias`~ No, it's just `Blend`,
    * ~`LinearVariableScaleBias`?~ No, it can be probably approximated with a `NormLPQ`, `Curve`s, and `Clamp`s. If `Curve` has overshoot/undershoot problems, add extra control points; a new`PiecewiseLinearCurve` module will probably not be necessary.
    * ~`Divide`?~ No, leave that for a different PR.
- [x] Support as much of RESTnoise module notation as possible:
  - [x] `"SourceModule" : "single_module"` (difficult due to polymorphism)
  - [x] `"Bounds" : [0.1, 0.9]`
  - [x] `"Angles" : [ 10.0, 25.5, -45.4 ]`
  - [x] `"Scale" : [ 2.0, 3.5, -1.4 ]`
  - [x] `"Scale" : 1.4`
  - [x] `"Translation" : [ 2.0, 3.5, -1.4 ]`
  - [x] `"NoiseQuality" : 2`
  - [x] `"Translation" : 4.3` (difficult to support both this and `"Translation" : [ 2.0, 3.5, -1.4 ]` due to polymorphism)
  * `"ControlPoints" : [ -0.5, 0.3, 1.1, 2.2 ]` for `Curve` modules? Or is that a bug? The [live example](http://restnoise.org:3001/networks/name:world2?gradient=realistic&bound0=-90&bound1=90&bound2=-180&bound3=180&maptype=sphere) uses the list-of-lists notation.
  * `"Roughness" : 3.3`? I think that's a bug. The live example uses ints.
  - [x] `"XDisplaceModule" : "mySourceModuleX", ...`
  * When alternative formats are supported, do not output module properties twice when encoding. Encoding modules is not yet an important feature; maybe leave it for another PR?
- [x] Test all module codecs, using the examples given in the RESTnoise docs ([link](https://restnoise.readme.io/docs/))
  - [x] Order module codec fields so that output JSON matches the canonical RESTnoise examples.
  - [x] Add example module directory as a test program option.
  - [x] Move test JSON strings into files rather than hardcoding them.
  - [x] Expand `EncodedModuleGroup` tests to ensure that all methods of setting source modules work correctly.
- [x] Use JSON-encoded module groups to generate tilesets.
  - [x] Add module group specifications to additional options files.
  - [x] Add module group locations to the options file.
  - [x] Decode module group locations from the options file.
  - [x] Make `ModuleGroup` a container rather than a `Module`? (Yes)
  - [x] Make module classes (with codecs) to replace functions like `makeQuadrantSelector`? They're combinations (simple or complex) of modules already existing, but in the RESTnoise module group format they would have to be written out in full, adding quite a lot of work to the user, and increasing the chance of making a mistake.
  - [x] Decode module groups and input/output module names from additional options files into `ModuleManager`.
  * ~Try to make an exception for input module names when setting source modules, because they need not be present in the module group? Alternatively, make a note that input module names should be bound to a dummy module which will be overwritten?~ (No. Use a `Forward` module)
  - [x] Make a `Forward` module for `ModuleGroup` inputs.
  - [x] Identify `QuadrantSelector` modules when decoding `ModuleGroup`s, and set their translation values each time the corner changes.
  - [x] Use decoded module groups in tile generation.
  - [x] Fix source modules being destructed.
  - [x] Use a single bool to turn on debug output for all modules. (temporary solution until #39 is fully implemented)
  - [x] #38 seems to have reappeared, fix it.
  * The output still doesn't look great, but I think tweaking/polishing the examples is a matter for a different pull request.
- [x] Miscellaneous:
  - [x] Clean up `EncodedModuleGroup::decode`/`ModuleWithSources::setModuleSources`.
  - [x] Verify that specified input modules have at least one source module?
  - [x] Make the module map and the quadrant selector list private in `ModuleGroup`, because messing with them can have bad effects. `EncodedModuleGroup` and its codec will still need access to them, though.
  - [x] Write JSON schema for Wangscape module group format.
  * ~Write JSON schema for RESTnoise module group format?~ I'm not sure that really needs to be done as part of Wangscape.
  - [x] Warn or fail when RESTnoise module notation allows the same attribute to be decoded from two different fields, and both are used? I can't see a simple way to do this in spotify-json; probably better to make it a condition in the module schemas.
  - [x] Remove `Reseedable*` classes.
  - [x] Move the map which converts module type names to decoding functions into ~its own file~ an anonymous namespace.
  - [x] Make a header like `noise/noise.h` that includes all useful noise classes.
  - [x] Add a default module group field to the options file, so that not every pair of terrains must be listed.
  - [x] Make a header that includes all the classes necessary to decode noise module groups (and keep it updated).
  - [x] Update CMakeLists.txt.
  - [x] Resolve merge conflicts (done, but there may be more when work is complete).
  - [x] Move `ModuleGroup::ModuleID` typedef into its own header and use it everywhere appropriate.


serin-delaunay:
For reference, the current procedure to decode a module group is approximately:
* Decode the object containing the module specifications to a `map<string, encoded_value_ref>`.
* For each item in the map:
  * Decode the module's `encoded_value_ref` to a `map<string, encoded_value_ref>`, and extract the value with key `"type"`. Decode that value to a `string` to find which module type to use.
  * Look up the module type string in a map to get a function which will convert the module's `encoded_value_ref` to a `ModulePtr` using the correct codec.
  * Use that function to decode the module's `encoded_value_ref` to a `ModulePtr` containing a `Wrapper<M>` of the correct type.
  * Decode the module's `encoded_value_ref` to a `NoiseSources` object containing the names of the source modules (this is done inside the above function).
  * Put the `ModulePtr` and `NoiseSources` in a map with the module's name as the key.
* For each `ModulePtr` and `NoiseSources`:
  * Set the `ModulePtr`'s source modules according to the named modules.
  * Put the `ModulePtr` in a map with the module's name as the key.
* Return that map.

serin-delaunay:
Regarding whether to remove the `Module` base class from `ModuleGroup`:

The current network of modules used in tile generation is inefficient because almost every module with sources points to a `ModuleGroup` containing the source modules instead of to the source modules themselves. This introduces an unnecessary `map<string, ...>` search between each layer of normal `Module::GetValue` calls.

These `ModuleGroup` interfaces are produced by building module networks using arithmetic operators and functions to build networks from `Reseedable`s without central ownership. Using the flat module network structure specified by RESTnoise, Most of these `ModuleGroup::GetValue` calls will be eliminated without any change to how `ModuleGroup`s work. Each terrain would have just three `ModuleGroup`s (vertical/horizontal border modules and central modules), and (for now) a single `ModuleGroup` would govern how they are combined to make a `TilePartition`. The total would be 16 `ModuleGroup::GetValue` calls per pixel per tile (with square tiles), which is much less of a concern.

So, the most important consideration in what to do with `ModuleGroup`s is the resulting API. They don't need to have a codec, as they will not provide any useful functionality to Wangscape users. Their use will be restricted to:
1. `ModuleManager`s store them.
2. `Raster*` classes request output values.
3. All modules in the group will have their seeds reset to different values.
4. Either a `ModuleGroup` or its output module will be set as another module's source.
5. Certain member modules will have their source modules changed regularly, and always to a `ModuleGroup` or its output module.

.

1. Makes no difference either way.
2. Favours keeping the `Module` base class, but it would be trivial to do `moduleGroup.getOutputModule()->getValue()`.
3. Can be done with the `Wrapper<M>`/`Reseedable<M>` interface, but perhaps it would make more sense as a distinct member method rather than an implementation of `Wrapper::setSeed`.
4. Makes a lot of sense with `ModuleGroup` as a subclass of `Module`, but any function that could take a reference to the unwrapped `ModuleGroup` as input could just as easily take the `ModuleGroup`'s unwrapped output module.
5. Is a nontrivial operation whether or not a `ModuleGroup` is a `Module`. There are two ways to do it:
  * Find the `ModulePtr` containing the module whose source must be changed, and call its `set*Module` method. This is made possible by the `WrapperBase` interface, which provides all three methods for all inner modules. These calls will be made repeatedly, so the program should store either the precise details of these calls (module name, source type, source index) or lambdas which will make those calls.
  * Modules with changeable sources point to a module which forwards the `GetValue` call to the correct input module. This module would have a common interface, so that only the module name needs to be stored, not the source type or index. This module could be a `ModuleGroup` (the output module name could be changed when necessary, or the module in the output slot could be overwritten). A `Cache` module would work just as well (call `setSourceModule(0, _)` to update it). The cache behaviour would be unnecessary, so a similar `Forward` module could be made for this purpose.

I favour removing the `Module` base class from `ModuleGroup`, and implementing `ModuleGroup` inputs using a `Forward` module which mimics `Cache`, but without the cache behaviour.

serin-delaunay:
I just realised that RESTnoise's `NoiseQuality` format is integers (presumably 0, 1, and 2), rather than the string enumeration (`"Fast"`, `"Standard"`, `"Best"`) I've been using so far. There are some other choices in the RESTnoise format which I disagree with or which may be difficult to implement with spotify-json:
* `Translation` field in `TranslatePoint` modules, which can be individual (`[1.1, 2.2, 3.3]`) or uniform (`4.4`). This matches the libnoise API, so it's probably best to implement it using a `one_of_t`.
* The same for `Scale` in `ScalePoint` modules.
* `SourceModule` field, which can be a single module (`"source_name"`) or an array (`["source_1", "source_2"]`). The array option should have a plural name (`"SourceModules" : ["source_1", "source_2"]`), and the singular option is not really necessary. This conflict is already causing some confusion in the Wangscape code, because the `SourceModule` field is decoded into a `vector<string>`, which internally has a plural name to reflect that.

There are a few options how to proceed:
1. Break RESTnoise compatibility and make a format which is based on RESTnoise but incompatible. I don't much like this option; I think every module network should have at least one format which is readable by both RESTnoise and Wangscape.
2. Implement at least one of RESTnoise's options, but provide alternatives which work only in Wangscape (`SourceModules` field, string enumeration for `NoiseQuality`). That way users can RESTnoise-compatible module groups if necessary, but otherwise are able to write `SourceModules` and `NoiseQuality` fields in a sensible way. Note that this will still require the `"SourceModule" : ["source_1", "source_2"]` format, because that is the only option in RESTnoise for modules like `Add`.
3. Implement a subset of the RESTnoise format, or the complete RESTnoise format. I don't like this option because of the `NoiseQuality` format; users must have a more readable alternative.

I favour option 2.

WangscapeTest/TestCornerCombiner.cpp
```diff
@@ -1,29 +1,30 @@
...
     noise::module::CornerCombinerBase cc05;
     noise::module::CornerCombinerBase cc2;
-    noise::Reseedable ccg2_p;
+    noise::module::ReseedablePtr ccg2_p;
     TestCornerCombiner():
-        cc1(), cc05(0.5), cc2(2.),
         ccg2_p(noise::module::makeCornerCombiner(true, true))
     {
+        cc05.SetPower(0.5);
+        cc2.SetPower(2.);
```
hryniuk: What are these coefficients? Could you name them?

WangscapeTest/TestModuleCodec.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+protected:
+    void remove_whitespace(std::string& s)
+    {
+        s.erase(std::remove_if(s.begin(),
+                               s.end(),
+                               [](char c) { return std::isspace(c, std::locale()); }),
+                s.end());
+    }
+};
```
hryniuk: It looks the same as in `WangscapeTest/TestEncodedModuleGroup.cpp`. Maybe just move it to a separate file 😃.

Wangscape/noise/module/NormLPQ.cpp
```diff
@@ -23,15 +22,38 @@ double NormLPQ::GetValue(double x, double y, double z) const
     return pow(dp, mQP);
 }
 
-void NormLPQ::setPQ(double p, double q)
+void NormLPQ::SetPQ(double p, double q)
+{
+    mP = p;
+    mQ = q;
```
serin-delaunay: Then `updateQP` would be called twice.

Wangscape/noise/module/codecs/CurveWrapperCodec.h
```diff
@@ -0,0 +1,47 @@
...
+                       [](const CurveWrapper& mw)
+                       {
+                           std::vector<std::pair<double, double>> all_control_points;
+                           const noise::module::ControlPoint* raw_control_points = mw.module.GetControlPointArray();
+                           const int control_point_count = mw.module.GetControlPointCount();
+                           for (int i = 0; i < control_point_count; i++)
+                           {
+                               const noise::module::ControlPoint control_point = raw_control_points[i];
+                               all_control_points.push_back({control_point.inputValue, control_point.outputValue});
+                           }
```
serin-delaunay: `std::transform needs the destination container pre-allocated. I would need to resize the vector first.

WangscapeTest/TestModuleCodec.cpp
serin-delaunay: It is the same! This is still a work in progress, after all ;)

Wangscape/noise/module/Wrapper.cpp
```diff
@@ -0,0 +1,67 @@
...
+        module.SetXDisplaceModule(source_module);
+        break;
+    case 1:
+        module.SetXDisplaceModule(source_module);
+        break;
+    case 2:
+        module.SetXDisplaceModule(source_module);
+        break;
+    default:
+        throw std::runtime_error("Tried to set displace module with index greater than 2");
```
serin-delaunay: Because two of those calls should be to `SetYDisplaceModule` and `SetZDisplaceModule`.

Wangscape/noise/module/Wrapper.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+template<> void Wrapper<Turbulence>::setSeed(int seed)
+{
+    module.SetSeed(seed);
+}
+
+template<> void Wrapper<Voronoi>::setSeed(int seed)
+{
+    module.SetSeed(seed);
+}
```
serin-delaunay: It's possible to use template magic to write one specialisation for all `M`with `SetSeed`, but for now the concise and readable version of that template isn't compatible with my compiler. And the number of modules with `SetSeed` isn't too large.

Wangscape/noise/module/Reseedable.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+template<> void Reseedable<Turbulence>::setSeed(int seed)
+{
+    module.SetSeed(seed);
+}
+
+template<> void Reseedable<Voronoi>::setSeed(int seed)
+{
+    module.SetSeed(seed);
+}
```
serin-delaunay: The libnoise classes with `SetSeed` don't implement it as a virtual function of the module base class (because there are modules that don't need a seed), so these are all "different" functions. At the moment, this is the simplest way to gather them all.

Wangscape/noise/module/ReseedableOps.cpp
```diff
@@ -0,0 +1,502 @@
...
+{
+    return left + -right;
+}
+
+ReseedablePtr operator-(ReseedablePtr left, double right)
+{
+    return left + -right;
+}
+
+ReseedablePtr operator-(double left, ReseedablePtr right)
```
serin-delaunay: That would only be possible for a `Reseedable` containing to a `Const` module, not the general case. In any case, all these operators will be deleted before this PR is merged.

Wangscape/noise/module/ReseedableOps.cpp
```diff
@@ -0,0 +1,502 @@
...
+}
+
+ReseedablePtr operator/(ReseedablePtr left, double right)
+{
+    return (left * (1. / right));
+}
+
+ReseedablePtr operator/(double left, ReseedablePtr right)
+{
+    throw std::runtime_error("Not implemented");
```
serin-delaunay: Static assert would probably be better, since this is never called. This was intended to be temporary, but now even if `Divide` modules are implemented this operator will be deleted.

Wangscape/noise/module/ReseedableOps.cpp
```diff
@@ -0,0 +1,502 @@
...
+{
+    auto max_p = std::make_shared<Reseedable<Max>>();
+    max_p->module.SetSourceModule(0, a->getModule());
+    max_p->module.SetSourceModule(1, b->getModule());
+
+    return makeModuleGroup({
+        {"source0", a},
+        {"source1", b},
+        {DEFAULT_MODULE_GROUP_OUT, max_p}
+    });
```
serin-delaunay: There's no reasonable way to compare two modules. This returns a `Max` module returning the pointwise greatest value of the two input modules. This function will be deleted anyway.

Wangscape/noise/module/ModulePtr.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <memory>
+#include "WrapperBase.h"
+
+namespace noise
+{
+namespace module
+{
+
+typedef std::shared_ptr<WrapperBase> ModulePtr;
```
serin-delaunay: Once the `Reseedable*` classes are removed, all modules used in Wangscape will be contained in a `Wrapper`. The exceptions will be:
*`RasterBase::build` takes a `Module&` parameter (this can be changed).
* `Module::SetSourceModule`, etc. take a `Module&` parameter.
* `Module::m_pSourceModule` internally uses a raw `Module**`.
I think that's enough to justify using the name `ModulePtr`, but it can be changed to `WrapperPtr` if necessary.

Wangscape/noise/ModuleManager.h
```diff
@@ -14,14 +14,15 @@ namespace noise
 class ModuleManager
 {
 public:
+    using ReseedablePtr = noise::module::ReseedablePtr;
```
serin-delaunay: Yes, I'm not sure why that's there. There's also `typedef noise::module::Module Module;`, which is totally unused in the class. I expect they're both relics from before the class was moved into the `noise` namespace.

Wangscape/noise/module/ReseedableOps.cpp
```diff
@@ -0,0 +1,502 @@
...
+    ReseedablePtr result = translate(norm_lp_q, -1., -1., 0.);
+    result = scalePoint(result, 2., 2., 0.);
+    result = scaleBias(result, 1. - min, min);
+    result = clamp(result, min, 1.);
+    return result;
+}
+
+ReseedablePtr makeVariableScaleBias(ReseedablePtr source, ReseedablePtr min, ReseedablePtr max)
+{
+    return ((source + 1)*0.5*(max - min)) + min;
```
serin-delaunay: These coefficients transform a `source` value in the range `[-1, 1]` to the range `[0, 1]`. It is then transformed to the range `[min, max]`.

I was considering making a module out of this so that users don't have to make 4 JSON modules to implement it. However, looking at the formula again, I think it's the same as `noise::module::Blend`, so that won't be necessary.

Wangscape/noise/module/codecs/CurveWrapperCodec.h
```diff
@@ -0,0 +1,47 @@
...
+                           return all_control_points;
+                       },
+                                      [](CurveWrapper& mw, std::vector<std::pair<double, double>> all_control_points)
```
serin-delaunay: I think you're right. I like to match multi-line argument lists with their opening parentheses, but when lambdas are in use that starts to become impractical.

Wangscape/noise/module/CornerCombinerBase.h
```diff
@@ -4,6 +4,8 @@ namespace noise
 {
 namespace module
 {
+// Default power for the noise::module::CornerCombinerBase module.
```
serin-delaunay: For classes derived from `noise::module::Module`, I'd prefer to imitate the libnoise coding style (hence the `PascalCase` method names). That includes comments on constant values whose purpose should be obvious. ([example](https://github.com/qknight/libnoise/blob/master/src/noise/module/rotatepoint.h#L45))

Wangscape/noise/module/ReseedableOps.cpp
```diff
@@ -0,0 +1,502 @@
...
+        {DEFAULT_MODULE_GROUP_OUT, invert_p}
+    });
+}
+
+ReseedablePtr& operator-=(ReseedablePtr& left, ReseedablePtr right)
+{
+    return (left = left + -right);
+}
+
+ReseedablePtr& operator-=(ReseedablePtr& left, double right)
```
serin-delaunay: In the old scheme for module networks, with `ModuleGroup`s containing `ReseedablePtr`s, there might have been some semantic differences between passing a `shared_ptr` and a `shared_ptr&`. But all of these operators will be deleted, so it's not worth worrying about.

Wangscape/noise/module/ReseedableOps.cpp
```diff
@@ -0,0 +1,502 @@
...
+}
+
+ReseedablePtr makeEdgeFavouringMask(double p, double q, double min)
+{
+    auto norm_lp_q = std::make_shared<Reseedable<NormLPQ>>();
+    norm_lp_q->module.SetPQ(p, q);
+    ReseedablePtr result = translate(norm_lp_q, -1., -1., 0.);
+    result = scalePoint(result, 2., 2., 0.);
+    result = scaleBias(result, 1. - min, min);
+    result = clamp(result, min, 1.);
```
serin-delaunay: The useful part of the texture generated by `NormLPQ` occupies the square `[-1, 1]×[-1, 1]`.The `translate` and `scalePoint` calls transform it into the square `[0, 1]×[0, 1]`.

The output values are mostly in the range `[0, 1]`, but some values of `p` and `q` will produce larger values. The `scaleBias` call transforms the `[0, 1]` range into `[min, 1]`, and the `clamp` call flattens any values outside of that (including any values which may be less than `min` due to floating point errors).

This is quite a complex operation, and I don't think Wangscape users should have to type out 5 JSON modules to implement it. I think it would be worthwhile to make this a real module.

WangscapeTest/TestCornerCombiner.cpp
serin-delaunay: `CornerCombiner` needs to be tested with `Power` parameters `< 1`, `== 1`, and `> 1`. These are just example parameters for which the expected output values can be easily calculated; they don't have any special meaning.

MarianSaeger:
Hi! 

I'm happy you find RESTnoise useful - please note that i tried to be as close to the libnoise API as possible, this is why for example SetSourceModule has no plural form. It was only a sideproject of mine, mainly to play around with SWIG. I find your thoughts very thorough and useful, i'd happy to let the ideas flow back into RESTnoise. =)

Best regards!

Wangscape/noise/module/CornerCombinerBase.h
hryniuk: Hmm... ok. In the example it looks like annotations of some documentation generation tool :thinking:.

Wangscape/noise/module/ModulePtr.h
hryniuk: Ok, `ModulePtr` seems reasonable 👍.

Wangscape/noise/module/NormLPQ.cpp
hryniuk: Yeah, I noticed. Never mind.

Wangscape/noise/module/CornerCombinerBase.h
serin-delaunay: Yep. I don't know what documentation tool Wangscape will use (probably doxygen), but when doc generation is added to the build process, comments like this will have to be formatted for compatibility.

serin-delaunay:
Hi Marian, thanks for getting in touch! I'm glad you find my ideas useful. I'm not using the RESTnoise interface, but its clear JSON format definitions have saved a lot of time for this project. This PR has a lot of text not related to RESTnoise, so it would probably be best to discuss possible API changes in a RESTnoise-dedicated space - perhaps [github issues](https://github.com/MarianSaeger/restnoise/issues) or [readme.io discussions](https://restnoise.readme.io/discuss)?

Best regards! :)

Wangscape/noise/module/Wrapper.cpp
hryniuk: Oh, great :)

Wangscape/noise/module/codecs/CurveWrapperCodec.h
hryniuk: And you've got `control_point_count`. Am I missing something?

serin-delaunay:
> Debug generated tiles (currently they're flat blends with black patches).

This is happening due to a memory error: while horizontal border module group for ["v","s"] is being decoded, the "OUTPUT" module in the combiner module group has its source module pointer reset to nullptr. Later (probably while source module pointers are being set for the new horizontal border module group), the same pointer is set to the perlin module from that new module group.

I guess the "OUTPUT" modules for both module groups somehow got the same `m_pSourceModules` pointer in their constructor. I can't think of any way that could happen, except for the Module::Module constructor not being called correctly. But both of those output modules are libnoise standard modules, which shouldn't have that problem, and I checked all the Wangscape module constructors.

I've switched Visual Studio builds to a self-built libnoise library with debug symbols; that should make this easier to fix. It seems the culprit is a`Forward` module, but I'm still not sure how it's causing a memory conflict. For some reason `Module::~Module` is being called during module decoding. I'm pretty sure that shouldn't happen; no modules are deleted until tile generation is complete.

serin-delaunay:
I'm currently writing the schema for Wangscape module groups, and there are some possible conflicts between:
* Indicating which modules get input from outside the group.
* Specifying the format of module source lists.
* Verifying that the source module relation on a module group forms a directed acyclic graph.

All libnoise and Wangscape modules have consistent numbers of source modules, and all of them should be specified to avoid error. So ideally the module group schema should check that the `SourceModule` and similar fields are present and contain the correct number of module names.

However, module groups (in particular the "combiner module group" used to build corner masks from central and border modules) can have input modules from outside the group, which are set dynamically before each use. The modules which have external sources must be types with at least one source module, so it seems that the module group schema should require them to have a standard-conforming `SourceModule` field, containing irrelevant data which will be overwritten before the module group is used.

At the moment, the module group/module codecs only check that each module named as a source module exists in the group. So the user can either leave the source list blank or name a real module (preferably either itself or a dummy module which will not be used).

When source lists' lengths are verified with a schema, it will not be possible for a single module type to perform that verification on its source list and allow blank source lists when used to get external input. The options to get around that are:
1. Users make a dummy module (preferably one that causes an error when used) to indicate source slots that will be used for external input. This will work in any situation, but is rather awkward and shouldn't be advocated as a default.
2. Modules with external input name themselves as the source that should be overwritten. If overwriting fails, Wangscape will enter an infinite recursive loop when trying to evaluate the module. I don't much like this solution; it should cause an immediate crash, and it should be much more obvious in the module group file. It will also be incompatible with any solution to #80.
3. Modules with external input use a special reserved name such as `""` or `"__EXTERNAL__"`. I prefer this to 2, but it doesn't make much sense to use special strings when JSON supports `null` as a basic value.
4. Modules with external input use a special non-string JSON value: `null`. This would be very clear, and compatible with #80, but would affect how noise source lists are parsed. I'd rather not add any more complication to module group decoding.
5. `Forward` modules' source lists are not verified (they can be blank or even missing). and if the user needs a different module to have external inputs (the only possible reasons are minute performance gains), they can use option 1. It seems weird to make one module type so strange and special, but it makes sense; this use case is exactly what `Forward` was made for. Since it doesn't do any processing, it's essentially useless in any other context, so not verifying its source modules (or verifying that it has no source modules) should not cause any problems.

I favour option 5.

serin-delaunay:
Everything on my to-do list is done, so I think this is ready for @hryniuk's review. :)

serin-delaunay:
The `json-noise-modules` branch diverged from `master` a long time ago (9/1/17), there have been several branches merged since then, and the conflicts are already resolved. Rebasing this branch on master introduces merge conflicts from the very first commit, and I don't think it's worthwhile to make all of that extra work. I'm just going to merge the new changes from master into this branch.

serin-delaunay:
It looks like I misconfigured CMake:
```
CMake Error at lib/CMakeLists.txt:2 (add_subdirectory):
  add_subdirectory given source "libnoise" which is not an existing
  directory.
```
I don't know how to fix that.

serin-delaunay:
Also, the travis script will have to be updated with the new example locations in `doc/`.

Wangscape/Options.h
```diff
@@ -17,13 +20,20 @@ class Options
     typedef std::vector<Clique> CliqueList;
 
     std::string filename;
+    std::string directory;
```
hryniuk: I'm thinking about extracting "meta data" of options from this class. What do you think?

I've looked over the all changes and they look good, I want to check the structure in detail this evening and we could merge it tomorrow.

serin-delaunay: That's probably a good idea. Where would you want to put them, considering that the values are used by functions that have access to `Options` but not `OptionsManager`?

hryniuk: If so, maybe the best idea is to define it inside Options, as public class, `Options::Location` or similar.

Wangscape/noise/ModuleManager.cpp
```diff
@@ -20,31 +72,72 @@ ModuleManager::ModuleManager(const Options & options) :
...
+                {
+                    if(options.defaultModuleGroup)
+                        mVerticalBorders.insert({tp, loadModuleGroup(p.string())});
+                    else
+                        throw std::runtime_error("Missing vertical module group, and no default module group");
+                }
             }
     }
+    
 }
```
hryniuk: Isn't this constructor too big? I seems that you can extract this `for` loops to a function (or a group of classes).

serin-delaunay: Yes, that would be better. I think the function/class will need two separate functions for default and specific module groups, and it'll need to be templated on the key datatype (`TerrainID` and `pair<TerrainID,TerrainID>` are both used).

Wangscape/Options.h
serin-delaunay: A class instance in `Options` sounds like the best idea. The code to update `directory` can be delegated there.

However, I'm not sure it makes sense for `OptionsLocation`/`Options::Location` for it to be a nested class when `TileFormat` and similar classes are independent?

hryniuk: I don't know either, which would be better in that case. You can make it standalone class for me :).

serin-delaunay:
OK, I've made the requested changes. I fixed some of the issues with CMake, but it's now failing with `/Wangscape/Wangscape/noise/EncodedModuleGroup.h:7:28: fatal error: spotify/json.hpp: No such file or directory`.

Wangscape/OptionsPaths.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+#include <string>
+class OptionsPaths
+{
+public:
+    virtual ~OptionsPaths();
+    void initialise(std::string filename);
```
hryniuk: Why you're not using a constructor?

serin-delaunay: `Options` stores an `OptionsPaths` value. The `Options` instance is constructed in a spotify-json function, and the `OptionsManager` can only initialise the `OptionsPaths` after the `Options` and `OptionsPaths` constructors have been called.

I could move the `initialise` code into a constructor, but it seems pointless to construct an `OptionsPaths` and then overwrite it with another one. It doesn't make much sense to wrap it in a `boost::optional` or `std::unique_ptr`, either.

I think it's worth making `ModuleGroupContainer::moduleGroupDescription` private and moving assignments ([link](https://github.com/serin-delaunay/Wangscape/blob/json-noise-modules/Wangscape/noise/ModuleManager.cpp#L14)) into a constructor, though.

hryniuk:
Ok, for me you can merge it.

Issue #69, opened 2017-01-09
Test AlphaCalculator and Raster classes


serin-delaunay:
They currently have no tests.
- [x] Test `AlphaCalculatorMax`
- [x] Test `AlphaCalculatorLinear`
- [x] Test `RasterImage`
- [x] Test `RasterValues`

PR #86, opened 2017-02-11
New alpha calculator: top two

serin-delaunay:
This adds a new method of alpha calculation, implementing point 4 in #37.

It has advantages from both `CalculatorMax` and `CalculatorLinear`:
* No incongruous black or bright patches.
* Power parameter allows transition sharpness to be controlled:
  * With a low power, large regions can blend terrains evenly.
  * With a high power, output will probably look like `CalculatorMax`.
  * With a moderate power, blended pixels are mostly confined to borders (the result looks like antialiased output from `CalculatorMax`).
* Calculations work correctly even with high powers.
* Calculations are simpler than CalculatorLinear.
* Ignoring the quantisation, I think the function is continuous everywhere except a region with dimension 1.

I added an optional field to the options file, to control the power parameter.

Since this calculator has output very similar to `CalculatorLinear` at low powers, and isn't affected by #51, it might be worth removing `CalculatorLinear` once this is merged.

TODO:
- [x] Add tests (#69).
- [x] Fix failing tests.
- [x] Add to CMakeLists.txt.
- [x] Resolve merge conflicts (after #103).


Wangscape/tilegen/alpha/CalculatorTopTwo.cpp
```diff
@@ -0,0 +1,54 @@
...
+void CalculatorTopTwo::updateAlphasAux(const Weights& weights)
+{
+    initialiseIndexedWeights(weights);
+    std::sort(mIndexedWeights.begin(), mIndexedWeights.end(), std::greater<IndexedWeight>());
+    const IndexedWeight& winner = mIndexedWeights[0];
+    const IndexedWeight& runner_up = mIndexedWeights[1];
+    const IndexedWeight& base = mIndexedWeights[2];
+
+    double gap_1 = winner.first - runner_up.first;
+    double gap_2 = runner_up.first - base.first;
```
hryniuk: What does it mean `gap_1` and `gap_2` :)?

serin-delaunay: They are the gaps between the winner and the runner-up, and between the runner-up and the base.   Perhaps `winner_margin` and `runner_up_margin` would be better?

hryniuk: Definitely they would be a bit more clear.

hryniuk: I wonder if they're needed. `total_gap` is equal to `winner.first - base.first` and it seems that `gap_1` and `gap_2` exist only for assertions. If this is the case, you could remove them, I think.

serin-delaunay: `gap_2` is used in `double runner_up_scaled = pow(gap_2 / total_gap, power);`. `gap_1` could be removed, if `assert(total_gap >= gap_2)` was added.

serin-delaunay:
Rebased on master.

Wangscape/tilegen/alpha/CalculatorTopTwo.h
```diff
@@ -0,0 +1,29 @@
...
+{
+
+
+class CalculatorTopTwo :
+    public tilegen::alpha::CalculatorBase
+{
+public:
+    double power = 1.;
+protected:
+    virtual void updateAlphasAux(const Weights& weights);
```
hryniuk: Mark it override, please.

serin-delaunay: Is that really necessary? `CalculatorBase::updateAlphasAux` is pure virtual, so `override` won't add any more safety in case of incorrect declarations.

#100's `CalculatorDither::updateAlphasAux` could use an `override` specifier, though.

hryniuk: Hmm... at first I've missed a fact that's it's virtual :D. Yes, it's safe now, but `CalculatorBase` doesn't look like a typical interface. That class is abstract now, but it has a few non-pure virtual functions, so I think it (its "abstracness") may change some day. With `override` keyword I would feel a bit safier. It's already marked `virtual`, which does less than `override`, cause it's just a hint for programmer. `override` will generate an error during compilation. For me it's a rule of thumb to use `override` every time it's possible.

serin-delaunay: You're right, some of the features of #37 will probably need the `CalculatorBase` class to change. Using `override` in classes derived from `CalculatorBase` could help to detect errors when fully implementing #37.

PR #90, opened 2017-02-21
Top left border modules (and significant bugfixes)

### LINKED ISSUE #52: Borders are too close to centre lines
Resolving #38 had the unexpected effect of making (almost?) all of the borders appear exactly halfway along the borders (example attached). Aside from being very undesirable in terms of variety, I suspect it's due to a fault in tile generation. At least it's a less ugly fault :)
s v

serin-delaunay:
Branched from #83. Implements option 8 in #52. Also fixed two significant bugs:
* Default border modules were not reseeded on load.
* `TilePartitionerNoise` used inconsistent corner numberings in `makePartition` and `makeCornerModule`.

### DIFF
--- a/Wangscape/Options.h
+++ b/Wangscape/Options.h
@@ -31,8 +31,8 @@ public:
     std::string combinerModuleGroup;
     boost::optional<std::string> defaultModuleGroup;
     std::vector<TerrainModuleGroupLocation> centralModuleGroups;
-    std::vector<BorderModuleGroupLocation> horizontalBorderModuleGroups;
-    std::vector<BorderModuleGroupLocation> verticalBorderModuleGroups;
+    std::vector<BorderModuleGroupLocation> leftBorderModuleGroups;
+    std::vector<BorderModuleGroupLocation> topBorderModuleGroups;
     bool debugOutput;
     
     virtual ~Options() = default;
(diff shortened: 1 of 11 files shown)


serin-delaunay:
The bug appears when a specific module group is used for a border. Tiles which use the default module group for all borders are unaffected. This happens even when the specific module groups are all loaded from the same file as the default module group. There seem to be two distinct "versions" of affected borders. Which one manifests on a given border depends on which terrain is in a corner unattached to the the border.

If a single left border uses a specific module, all horizontal borders can be affected (but not vertical borders), even those which don't connect the two terrains for which a module was specified.

A possible source of the difference in behaviour seems to be that default border modules didn't have seeds set when loaded. Now their seeds are set. However, making default border modules behave more like specific border modules doesn't seem likely to resolve a bug that appears only when using specific border modules.

serin-delaunay:
This can be rebased on #83 when its CMake errors are fixed and its review is complete, or on master after #83 is merged.

serin-delaunay:
Somehow resolving merge conflicts caused two submodules to be deleted. :S

serin-delaunay:
This is building and passing tests on Travis (at least up to #95), so it's ready for review/merge.

Issue #32, opened 2016-12-14
Support non-square tile geometries


serin-delaunay:
Currently Wangscape only supports writing square tiles.
- [ ] Support rectangular tiles.
- [ ] Support non-rectangular four-cornered tiles (using custom replacement for the corner-combiner and edge-favouring masks). Rhombic tiles will be the priority here, but an ideal solution would support any base tile mask that tessellates like a parallelogram.
- [ ] Support hexagonal tiles. The output corner Wang tiles will be triangles, but they must be built for 2 different orientations. The tile count per clique will be 2*n^3 (rather than n^4 for rectangular tiles). This may require the support for user-defined masks for the base tile shape, corner-combiners, edge-favouring masks, etc.
- [ ] Support non-hexagonal six-cornered tiles. Probably using the same framework to the non-rectangular four-cornered tiles.
- [ ] Support triangular and three-cornered non-triangular tiles. The output corner Wang tiles will be hexagons, and the tile count per clique will be n^6. This may be impractical, especially since games with triangular tiles are so rare. As such, this is a low-priority task.

serin-delaunay:
Some notes about how to add support for non-rectangular four-cornered tiles (with a view to implementing three- and six-cornered tiles in the future):

* Tile resolution becomes a bounding box rather than defining tile geometry.
* Tile geometry should be specified using a bitmap mask (two masks for triangular tiles). This mask should have a black or transparent background, and the region associated with each corner should have a distinct colour.
* An "anchor" pixel should be specified (the centre or top-left of a display tile.
* Two offset vectors should be specified, defining how the tile tessellates (in six-cornered tiles the third offset vector will be the sum or difference of the other two).
* Test that the mask(s) has/have no overlaps or gaps when tessellated according to the offset vectors. The mask and each of its corner regions should be connected, to make this test easier.
* The dual mask (two tiles for hex tiles) is/are found by recombining the corner regions according to the offset vectors.
* The dual mask(s) and the original mask(s) may have different bounding boxes. The output tilesets will be based on the dual mask(s).
* Generating corner masks for output tiles:
  * Quadrant selector module translations must be based on the offset vectors.
  * Three of the modules modules in the [default combiner module group](https://github.com/Wangscape/Wangscape/blob/master/doc/examples/example3/combiner_module_group.json) must be replaced:
    * The corner combiner (LL27-45)
    * The corner distance and peaks (LL51-99)
    * The edge-favouring mask (LL105-130)
[...]

PR #91, opened 2017-02-21
Make resolution option a vector; support odd values

### LINKED ISSUE #12: Support odd tile resolutions
At the moment various parts of the code assume the tile resolution is even. Odd resolutions have not been tested. There may be unforeseen errors or crashes when the tile resolution is odd.

The correct behaviour of #11 in particular is unclear - perhaps it should be an option in the JSON options.

serin-delaunay:
Resolves #12. Resolves item 1 from #32.

Update everything that uses resolution, including tile partitioners,
tile[set] generator, metaoutput, terrain image rearranger, examples.

Apart from changing the resolution option from an integer to an array of two integers, there is another important and less obvious change to the options file format: source tile offsets in images are now counted in pixels, not tiles. This doesn't affect any of the current examples, which have offset `[0,0]` for all source terrain tiles.

An eyeball test with resolution `[31, 29]` indicates terrain images are
correctly rearranged even with odd resolutions, but it's probably worth
making a real test case.

Needs matching changes in Wangview (https://github.com/Wangscape/Wangview/pull/14), which should be merged at the same time.


serin-delaunay:
Rebased on master.

Wangscape/tilegen/partition/TilePartitionerGradient.cpp
```diff
@@ -21,22 +22,22 @@ void TilePartitionerGradient::makePartition(TilePartition & regions,
...
+        for (size_t y = 0; y < mOptions.tileFormat.resolution.y; y++)
         {
-            weights[0] = static_cast<double>(gradientWeight(x, y, 0, 0, quarter_res));
-            weights[1] = static_cast<double>(gradientWeight(x, y, 0, resolution_sub_1, quarter_res));
-            weights[2] = static_cast<double>(gradientWeight(x, y, resolution_sub_1, 0, quarter_res));
-            weights[3] = static_cast<double>(gradientWeight(x, y, resolution_sub_1, resolution_sub_1, quarter_res));
+            weights[0] = static_cast<double>(gradientWeight({x, y}, {0, 0}, quarter_res));
+            weights[1] = static_cast<double>(gradientWeight({x, y}, {0, resolution_sub_1.y}, quarter_res));
+            weights[2] = static_cast<double>(gradientWeight({x, y}, {resolution_sub_1.x, 0}, quarter_res));
+            weights[3] = static_cast<double>(gradientWeight({x, y}, {resolution_sub_1.x, resolution_sub_1.y}, quarter_res));
```
serin-delaunay: These four `weights[]` assignments produce warnings on 64-bit gcc because a `size_t` (unsigned long) is implicitly converted to an `unsigned int`.

serin-delaunay: Fixed.

Wangscape/TerrainImages.cpp
```diff
@@ -45,10 +46,8 @@ void TerrainImages::addTerrain(TerrainID tid, std::string image_filename, std::s
         it = source_images.insert({ filename, img }).first;
     }
     sf::Vector2i offset(offset_x, offset_y);
-    sf::Vector2i box(resolution, resolution);
-    offset *= (int)resolution;
```
hryniuk: Why is this line removed?

serin-delaunay: These two lines caused the source tile offset in the source image to be measured in multiples of the resolution instead of in pixels. It didn't seem like a useful restriction; eventually some users may have source images with tiles not aligned to the grid.

PR #94, opened 2017-02-21
Add visual examples to front page README.md

### LINKED ISSUE #42: Add examples and diagrams to documentation
Now that the program actually generates corner Wang tiles, it would be good to illustrate what they look like and how they should be used, either in README.md or on a wiki page. This is an image generating tool, and it should be described with images.

serin-delaunay:
Resolves #42.

Mentions features from #83, so #83 should probably be merged first (even though this is branched from master).


serin-delaunay:
This does not and will not make any changes to executable code, so I don't think it needs rebasing.

hryniuk:
This is **awesome** :clap::exclamation:. Checking #83 now 😄.

serin-delaunay:
The squares sample:
wangscape-squares-sample
should be altered to use the same terrain map as the "natural" example:
wangscape-natural-sample

Issue #102, opened 2017-03-01
CMake builds are broken


serin-delaunay:
Introduced in #83.

Issue #99, opened 2017-02-25
Travis script doesn't find options file for testing


serin-delaunay:
See for instance https://travis-ci.org/Wangscape/Wangscape/builds/204860741:

> [Using options file at Wangscape/example/example_options.json](https://travis-ci.org/Wangscape/Wangscape/builds/204860741#L442)

...

> [unknown file: Failure
C++ exception with description "Could not open options file" thrown in the test fixture's constructor.
[  FAILED  ] TestMetaOutput.TestMetaOutputCorrect (0 ms)](https://travis-ci.org/Wangscape/Wangscape/builds/204860741#L477)

It looks like `.travis.yml` calls `./bin/WangscapeTest` without any command line arguments, so altering that call should be enough. I'm not sure what directory the call is made from.

The test options moved from `Wangscape/example/example_options.json` to `doc/examples/example2/example_options.json` in #83, so if this is resolved before #83 is merged then #83 will break it again.

Issue #95, opened 2017-02-21
Wangscape depends on display


hryniuk:
While running in a Docker container, I'm getting this error:
```
Failed to open X11 display; make sure the DISPLAY environment variable is set correctly
Aborted (core dumped)
```

I don't see a need for using a display as Wangscape doesn't display anything, it only generates some images. I suppose there's a different way of using SFML. If no, we could use some different, maybe smaller library.

serin-delaunay:
Hmm. Wangscape doesn't use any classes from `SFML/Window.hpp`, so it shouldn't be trying to open an X11 display. It does use OpenGL methods and `RenderTexture`s to assemble tilesets; perhaps they require X11?

There have been a few reports of this error; [one](http://en.sfml-dev.org/forums/index.php?topic=20680.0) was solved by not running the program as root; the [other](https://github.com/SFML/SFML/issues/508) caused an in-depth discussion but I can't see where the poster's issue was resolved.

If running as non-root doesn't fix the problem, could you get a stack trace to find where in the Wangscape code the error occurs? Setting a breakpoint at https://github.com/SFML/SFML/blob/master/src/SFML/Window/Unix/Display.cpp#L60 should do it, I think. If it's something that can be replaced, and there's no way to start an X11 instance in the docker environment, then Wangscape will have to either:
* Replace SFML with a non-GL dependent library (and probably rewrite the image collage routines as loops), or
* Clearly note that Wangscape can't be used in Linux environments without X11.

Boost GIL is a possible library for image I/O, although when I tried to use it in initial versions of Wangscape I found the API rather unwieldy. One alternative would be png++.

serin-delaunay:
A temporary workaround would be to have the travis script skip all the tests which include the offending function call. http://stackoverflow.com/a/7208148 would be one way to do this.

serin-delaunay:
[Xvfb](https://www.x.org/releases/X11R7.6/doc/man/man1/Xvfb.1.xhtml) may provide a permanent solution if properly configured ([example](https://github.com/travis-ci/travis-ci/issues/6987#issuecomment-266754776)).

serin-delaunay:
I've made #119 as a temporary workaround until Travis tests run with Xvfb.

PR #103, opened 2017-03-01
Fix linux build

hryniuk:


serin-delaunay:
Travis builds now reach the linking stage of Wangscape before failing. Looks like CMake needs to tell Wangscape how to access `double-conversion`.

serin-delaunay:
OK, it's building and running on Travis, and passing all tests until it triggers issue #95. That's beyond the scope of this pull request, so I'm merging.

This resolves #102 and #99.

Issue #93, opened 2017-02-21
Replace square and gradient tile partitioners with module group-based configurations


serin-delaunay:
I'm using `TilePartitionerSquare` to generate examples of "non-Wangscape" tilesets for #42. At the moment I have to recompile Wangscape to do that; setting up module groups to produce square tiles ~seems to be a fool's errand~ takes at least a little bit of thought.

~It would be nice to select the tile partitioner type using a text field/enum in the configuration file, like what #86 introduces for the alpha calculator.~
- [x] Make a module-based Wangscape configuration example that imitates `TilePartitionerSquares`. (#108)
- [x] Make a module-based Wangscape configuration example that imitates `TilePartitionerGradient`. (#130)
- [x] Remove `TilePartitionerGradient`.
* `TilePartitionerSquares` should be left intact because it makes some of the testing code much simpler. It could be moved into `WangscapeTest` or removed as part of #26.

serin-delaunay:
Alternatively, make simple example module groups to replace the non-noise tile partitioners.

PR #108, opened 2017-03-08
Add new example: squares

serin-delaunay:
Useful for generating Wangview-compatible tilesets which look like unprocessed source tiles, or as a base for other tilesets with non-noisy boundaries.

Will deprecate ~`TilePartitionerPerlin`~`TilePartitionerSquares` and affect #93.

I don't know why the combiner module group needs:
```json
    "and_t":{
      "type": "TranslatePoint",
      "Translation": [-0.5, 0.5, 0],
      "SourceModule":"and"
    },
```
I would have expected [0.5, 0.5, 0], because `and` is symmetric in `x=y`, but that causes the corner masks to be flipped horizontally. Either there's a fault in corner mask generation that #90 didn't fix, or I'm misunderstanding the module group I wrote.

Tested in #90 before committing, but branched from master because it probably doesn't depend on the changes in #90.


hryniuk:
After rebase on `master`, running:
```shell
./Wangscape example_squares/example_options.json 
```

causes

```shell
terminate called after throwing an instance of 'spotify::json::decode_exception'
  what():  Unexpected input
spotify::json::decode_exception encountered at 216: Unexpected input[1]    14413 abort (core dumped)  ./Wangscape example_squares/example_options.json
```

Could you check it?

hryniuk:
I wonder, if we could use images (`sea.png` and `grass.png`) that are already in `Wangscape`'s repository, setting proper path in options file. This will save us from making the repository bigger, I think.

serin-delaunay:
Yes, it would make sense to put a single copy of each source tile in a folder like `doc/images/tiles`. Not in `doc/examples/tiles`, because the AppVeyor scripts would try to run Wangscape on `doc/examples/tiles/example_options.json`. I would prefer to do that in a separate pull request, though.

I'm checking how `example_squares` runs after rebasing the branch now. Do you have a Wangscape executable in `doc/examples`?

serin-delaunay:
[Wangcheck](https://github.com/Wangscape/Wangcheck) found an error:
```
Failed validating 'type' in schema['properties']['TileFormat']['properties']['Re
solution']:
    {'description': 'Specifies the width and height of the square tiles '
                    'Wangscape will operate on.',
     'items': {'maximum': 512,
               'minimum': 8,
               'multipleOf': 1,
               'type': 'integer'},
     'maxItems': 2,
     'minItems': 2,
     'title': 'Tile resolution',
     'type': 'array'}

On instance['TileFormat']['Resolution']:
    32
```
The example wasn't updated when #91 was merged. I've fixed that, and there are no more complaints from Wangcheck, but I'm still getting a crash.

serin-delaunay:
Debug output was enabled. Without the fixes in #111, Wangscape crashes if there isn't a `test` folder in the current directory. I've turned debug output off, which fixes it.

Even with the fixes in #111, the debug output routine waits for `stdin`. Until we have a non-interactive way to limit the number of debug tiles produced, examples should never be committed with debug output enabled.

PR #111, opened 2017-03-24
Improve debug output

### LINKED ISSUE #109: Output debug data in a more sensible way
Currently #39 is unimplemented, and instead Wangscape has an option to output debug data for one corner at a time. The output is saved to a folder called "test" in the current working directory, and fails if the folder doesn't exist. The debug images have size 256x256, regardless of the tile resolution. They are overwritten for each corner.

- [x] Put the debug tiles in `output/debug`, not `cwd/test`.
- [x] Automatically create the debug folder.
- [x] Add debug resolution option to options.
- [x] Add option to use a directory structure to avoid/reduce overwriting debug tiles (potentially consuming a lot of disk space).

#39 will require some significant restructuring of the tileset generation code, and fixing these issues will be much quicker and simpler. I think it's worthwhile to postpone #39 for another milestone and make this part of https://github.com/Wangscape/Wangscape/milestone/1.

serin-delaunay:
Resolves #109.

Debug data is written to `output/debug/t.t.t.t.CC/`, where each `t` is the ID of a terrain type and `CC` identifies the active corner (`[TB][LR]`). This doesn't eliminate overwriting; two tiles in different tilesets with the same terrains in the corners will have debug data written to the same four folders.

The code to generate the `t.t.t.t` notation has some functionality in common with `TilesetGenerator::getOutputImageFilename`. I'm not sure whether it's worth unifying their implementations; the two functions perform this operation on two different data types (which just happen to be `typedefs` of a common type, they both apply different operations to the result, and the operation is really simple.


doc/schemas/options_schema.json
```diff
@@ -301,6 +313,7 @@
     "Terrains",
     "Cliques",
     "AlphaCalculatorMode",
-    "CombinerModuleGroup"
+    "CombinerModuleGroup",
+    "DebugTileResolution"
```
hryniuk: Should it be **required**?

doc/schemas/options_schema.json
```diff
@@ -293,6 +293,18 @@
...
+      "type": "array",
+      "title": "DebugTileResolution schema",
+      "description": "Specifies the resolution of the images written as debug data during tile generation.",
+      "items": {
+        "type": "integer",
+        "multipleOf": 1,
+        "minimum": 1
+      },
+      "minItems": 2,
+      "maxItems": 2
```
hryniuk: Same thing here: could it be `{"width": w, "height: h}`?

I think it will be nice to stick to one resolution format to use everywhere (it could be an array, if you think it's sufficient).

doc/schemas/options_schema.json
serin-delaunay: Hmm... probably not. I'll see if I can get spotify-json to use a sensible default. I've been using `(256,256) as the default debug tile size for `(32,32)` tiles; maybe `tile size * (8,8)` would be a good default.

Wangscape/Options.h
```diff
@@ -35,6 +35,7 @@ class Options
     std::vector<BorderModuleGroupLocation> leftBorderModuleGroups;
     std::vector<BorderModuleGroupLocation> topBorderModuleGroups;
     bool debugOutput;
+    std::pair<size_t, size_t> debugTileResolution;
```
hryniuk: Have we a type for "resolution"? Could we have one :)?

serin-delaunay: We don't. There's an issue about that: #46.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
```diff
@@ -43,6 +44,19 @@ noise::module::ModulePtr TilePartitionerNoise::makeCornerModule(const Corners& c
     combiner.setInputModuleSource(2, central.getOutputModule());
     if (mDebugOutput)
```
hryniuk: Code under this `if` should be moved to a separate function in my opinion. Notice that it's now longer than the action above, for which it saves debug info.

Hmm... Can it be moved to a separate module? Will there be any benefit? Can it be reused?

Wangscape/Options.h
hryniuk: Oh, opened in December. I have to bring more code to Wangscape to get rid of all these issues.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
serin-delaunay: We can move it to a separate function, but the function would have `corners, left, top, central, border_h, border_v, combiner` as parameters. Splitting it up might be more feasible. `write_map` and `write_group` could be moved into private class methods. The same can be done for the folder/filename generators and the console IO. The block would probably manageable after that.

I don't think it's worthwhile moving this into a separate module; this is a temporary solution until #39 is implemented.

doc/schemas/options_schema.json
serin-delaunay: Done. User can specify an exact resolution, otherwise it defaults to `tile size * (8,8)`.

Perhaps the user should just specify a multiplier; in that case it should probably be a separate issue.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
serin-delaunay: Done. Debug output required a lot of persistent objects, so I've made a class to do it.

serin-delaunay:
I think this is ready to merge. There are no merge conflicts, but if necessary I can pull changes from `master` to get the CI builds working.

Wangscape/Wangscape.vcxproj.filters.orig
```diff
@@ -0,0 +1,191 @@
...
+    <ClInclude Include="noise\WangscapeNoise.h" />
+    <ClInclude Include="noise\WangscapeNoiseCodecs.h" />
+    <ClInclude Include="noise\ModuleGroupContainer.h" />
+    <ClInclude Include="OptionsPaths.h" />
+    <ClInclude Include="codecs\Vector2Codec.h" />
+<<<<<<< Updated upstream
+=======
+    <ClInclude Include="DebugWriter.h" />
+    <ClInclude Include="tilegen\CornerTerrains.h" />
+>>>>>>> Stashed changes
```
serin-delaunay: Yep, this is noise from kdiff3. I don't like that it leaves these files around. I'll add `*.orig` to `.gitignore`.

hryniuk:
I've run it, it seems to be really helpful :).

PR #112, opened 2017-03-24
Update options schema with rectangular resolution

serin-delaunay:


doc/schemas/options_schema.json
```diff
@@ -48,12 +48,17 @@
       "additionalProperties": false,
       "properties": {
         "Resolution": {
-          "type": "integer",
+          "type": "array",
```
hryniuk: Could we have an object with `width` and `height` fields here? I think it will be clear and it will ease debugging/checking with Wangcheck/interpreting these options.

serin-delaunay: That's definitely an idea we should consider, but I think it should be a separate issue/PR. I forgot to update the schema in #91, and this PR is just fixes that. Changing the options format would also mean updating the `TileFormat` codec and example options files.

PR #119, opened 2017-03-28
Travis: Exclude tests that require X11

serin-delaunay:
As a workaround for #95, makes WangscapeTest skip test matching the following patterns in Travis:
* `*TestMetaOutput*`
* `*Squares*`
* `*TestGradientPartition*`
* `*TestTilesetGenerator*`

Currently they cause the test executable to crash, so any tests after the failed test don't get run. These can be reinstated when Travis tests run with Xvfb.

`TestTilePartitionerGradient.TestGradientWeight` fails due to a legitimate bug (missing `std::min` call), which is fixed in #116. #116 doesn't yet address everything in #20; it's missing packaging instructions. Implementing packaging could be done in a different PR, if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ services:
 script:
     - pwd
     - mkdir build
-    - docker run -v $(pwd):/Wangscape lukequaint/wangscape-build:latest /bin/sh -c "cd /Wangscape/build && cmake .. && make && ./bin/WangscapeTest ../doc --gtest_filter=-*TestMetaOutput*:*Squares*:*TestGradientPartition*"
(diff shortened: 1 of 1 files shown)


hryniuk:
One test is failing now :confused:.

serin-delaunay:
Yes. That test has a fix in #116. There are 3 options:
* Apply the fix in this branch too, and cause a merge conflict.
* Merge both this and #116 now, and implement the remaining tasks of #20 in a different PR.
* Implement all of #20 in #116 before merging. That means figuring out how to run Python on Appveyor.

serin-delaunay:
#125 fixed all these tests. Closing.

PR #125, opened 2017-04-29
Add Xvfb and fix Travis build

### LINKED ISSUE #95: Wangscape depends on display

hryniuk:
Fixed #95 by providing xvfb and configuration of a virtual display.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,8 @@ services:
 script:
     - pwd
     - mkdir build
-    - docker run -v $(pwd):/Wangscape lukequaint/wangscape-build:latest /bin/sh -c "cd /Wangscape/build && cmake .. && make && ./bin/WangscapeTest ../doc"
+    - docker run -v $(pwd):/Wangscape lukequaint/wangscape-build:latest /bin/sh -c \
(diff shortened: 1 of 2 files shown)


hryniuk:
@serin-delaunay all checks have passed :tada: 🎆 

This PR is connected to that one in the dockerfile's repository:
https://github.com/Wangscape/wangscape-build-dockerfile/pull/2

I've added these two lines:
```
Xvfb :1 -screen 0 1600x900x24 &> xvfb.log &
DISPLAY=:1.0; export DISPLAY;
```
to `.travis.yml`, because I think a display configuration should be as close as possible to running `xvfb-run` (it uses the same display's settings). It could be placed in a separate script or even (I'm not sure, if it's possible or recommended) in a `Dockerfile`, but I think that wouldn't be as clear as it is now.

serin-delaunay:
Awesome work!

PR #130, opened 2017-05-03
Add new example: gradient

serin-delaunay:
Makes `TilePartitionerGradient` redundant. See also #108. Affects #93.


hryniuk:
As in #108 I'm wondering, if we could reuse images that are now in `doc/` directory and if it makes sense at all (it will make harder copying these examples and running `Wangscape` with them from different directory, e.g. `build`). What do you think?

serin-delaunay:
I think it's worth discussing. Could you open a separate issue for it?

hryniuk:
Ok.

PR #133, opened 2017-05-05
Test rasters and alpha calculators

### LINKED ISSUE #69: Test AlphaCalculator and Raster classes

serin-delaunay:
Resolves #69.


WangscapeTest/TestCalculatorLinear.cpp
```diff
@@ -0,0 +1,93 @@
...
+    ptrdiff_t validateResults(const tilegen::alpha::CalculatorBase::Weights& weights)
+    {
+        cl.updateAlphas(weights);
+        const auto& alphas = cl.getAlphas();
+        int total_alpha = std::accumulate(alphas.cbegin(), alphas.cend(), 0);
+        EXPECT_EQ(255, total_alpha);
+        double total_weight = std::accumulate(weights.cbegin(), weights.cend(), 0.);
+        for (size_t i = 0; i < 4; i++)
+        {
+            double alpha_fraction = double(alphas[i]) / 255;
```
hryniuk: Why not `total_alpha`? It would make both alphas and weights symmetric.

WangscapeTest/TestCalculatorLinear.cpp
```diff
@@ -0,0 +1,93 @@
...
+            double weight_fraction = double(weights[i]) / total_weight;
+            EXPECT_NEAR(alpha_fraction, weight_fraction, 1.5 / 255);
```
hryniuk: Why does this "epsilon" equal `1.5 / 255.0`? Is it an arbitrary value or has it some not so obvious justification?

WangscapeTest/TestCalculatorMax.cpp
```diff
@@ -0,0 +1,81 @@
...
+    TestCalculatorMax()
+    {
+        std::random_device rd;
+        rng.seed(rd());
+    }
+    int countValue(const tilegen::alpha::CalculatorBase::Alphas& alphas, int value)
+    {
+        return std::accumulate(alphas.cbegin(), alphas.cend(), 0,
+                               [&value](int acc, sf::Uint8 x)
+        { return (x == value) ? acc + 1 : acc; });
```
hryniuk: I think `std::count` is what you're looking for :wink:.

WangscapeTest/TestRasterImage.cpp
```diff
@@ -0,0 +1,105 @@
...
+protected:
+    noise::module::GradientX gradientXModule;
+    noise::module::GradientY gradientYModule;
+    noise::module::Const constModule;
+    sf::Image image;
+    TestRasterImage()
+    {
+        constModule.SetConstValue(0.);
+    };
+    ~TestRasterImage() {};
```
hryniuk: Do you need a destructor here? If yes (nothing comes to my mind 😕), use `= default` to keep the code consistent.

WangscapeTest/TestRasterImage.cpp
```diff
@@ -0,0 +1,105 @@
...
+    EXPECT_EQ(1., ri.bounds().height);
+    for (size_t x = 0; x < 20; x++)
+    {
+        for (size_t y = 0; y < 20; y++)
+        {
+            EXPECT_EQ(0, ri.get(x, y).r);
+            EXPECT_EQ(0, ri.get(x, y).g);
+            EXPECT_EQ(0, ri.get(x, y).b);
+        }
+    }
```
hryniuk: Save a size, `20`, in a variable and use the name in these loops.

WangscapeTest/TestCalculatorMax.cpp
serin-delaunay: I think it's `std::count_if`, but yep.

WangscapeTest/TestCalculatorLinear.cpp
serin-delaunay: Alpha values returned by the `alpha::CalculatorBase` are quantised as integers from 0 to 255, producing an error up to `0.5/255`. When the initial division of alpha points has a remainder, [up to 1 point can be added to each corner](https://github.com/serin-delaunay/Wangscape/blob/master/Wangscape/tilegen/alpha/CalculatorLinear.cpp#L57), increasing the maximum error to `1.5/255`. I'll add a comment explaining that.

WangscapeTest/TestRasterImage.cpp
serin-delaunay: Looks like it probably came from copying `TestCornerCombiner`. I'll check if there are any other test classes that need a destructor removed or `default`ed.

WangscapeTest/TestCalculatorLinear.cpp
serin-delaunay: I could. It would make the `EXPECT_NEAR` test more accurate when `total_alpha != 255`, but it would also mean checking and skipping that test when `total_alpha == 0`. I think this would be better:
```cpp
            double alpha_fraction = double(alphas[i]);
            double weight_fraction = double(weights[i] * total_alpha) / total_weight;
            EXPECT_NEAR(alpha_fraction, weight_fraction, 1.5);
```

WangscapeTest/TestCalculatorMax.cpp
serin-delaunay: There are some `std::accumulate` calls already in `master` which could be replaced with `count_if`, and I think the `CartesianPower` tests have a for loop that could be replaced by `count`. I think that's a matter for another PR, though.

WangscapeTest/TestRasterImage.cpp
serin-delaunay: Done.

PR #135, opened 2017-05-21
Logging

### LINKED ISSUE #29: Logging
It will be nice to have a logger with a command line option, e.g. `-v` to print (to stdin or a file) what's going on. After quick googling I think for Wangscape it will be the best to roll its own small logger.

hryniuk:
Fixes #29


hryniuk:
I've been trying to find a balance between flexibility and simplicity of `logging` module. As Wangscape doesn't run for a long and only reads input/writes output, I didn't include printing e.g. time. On the other hand, it will be quite easy to add things like a file and a line the given line was logged from by extending `LOG_` defines, without changing logs that are already in a code.

hryniuk:
Two things to do for me:
- [ ] fix appveyor build
- [ ] review code style

serin-delaunay:
Excellent. I'll take a look later today.
Regarding including times in logs: Wangscape is currently quite slow (try an example with a clique of size 3 or 4 to see how slow). Currently, adding features is more important than optimising performance, but when we start to work on improving performance, log timestamps will be useful. That functionality can be added later, though.

hryniuk:
I'm not sure if it should be done by logs. The fact that they're printed on, for example, stdout, makes Wangscape a bit slower and will impact timestamps.

It would be better to have some performance tests to check results and think about profiling in the future. I've done some profiling with `valgrind` in LibreOffice (`callgrind` in fact) before and I plan to learn how to use `perf` (http://www.brendangregg.com/perf.html) - after that I will be able to profile Wangscape in Linux :).

Wangscape/logging/Logging.h
```diff
@@ -0,0 +1,21 @@
...
+#include "AppenderType.h"
+#include "Level.h"
+#include "Logger.h"
+
+#include <ostream>
+
+#define LOG_ERROR getLogger(logging::Level::Error)
+#define LOG_WARNING getLogger(logging::Level::Warning)
+#define LOG_DEBUG getLogger(logging::Level::Debug)
+#define LOG_INFO getLogger(logging::Level::Info)
```
serin-delaunay: I would prefer to avoid using `#define` unless it's absolutely necessary. Could these macros be replaced with functions or `static` (`const`?) instances/references?

I see that adding filenames and line numbers to messages would be easier with a `#define`, but that could be done with a `FILE_LINE` macro and `LOG_ERROR << FILE_LINE << ": Error message\n";`.

I don't think it would make sense if that was a standard part of `LOG_*`, or even `LOG_ERROR`. Many errors will be caused by faulty configurations, and the error message should direct attention to the code that needs fixing.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
```diff
@@ -45,6 +46,7 @@ noise::module::ModulePtr TilePartitionerNoise::makeCornerModule(const Corners& c
     combiner.setInputModuleSource(2, central.getOutputModule());
     if (mDebugOutput)
     {
+        setLoggingLevel(logging::Level::Debug);
```
serin-delaunay: Logging level and append mode should be set by the user either in command line options or in the JSON config file, and this change to the logging level should be undone before leaving this code block.

Also, there should be a comment explaining why the logging level is changed. I guess it's because the messages below are required to prompt user interaction.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
```diff
@@ -114,18 +116,18 @@ void TilePartitionerNoise::writeDebugData(const noise::ModuleGroup & central,
...
 
-    std::cout << "Writing debug modules for " << mDebugWriter->getCornerDescription() << "...\n" ;
+    LOG_DEBUG << "Writing debug modules for " << mDebugWriter->getCornerDescription() << "...\n" ;
     mDebugWriter->writeDebugGroup(central, "central");
     mDebugWriter->writeDebugGroup(border_h, "border_h");
     mDebugWriter->writeDebugGroup(border_v, "border_v");
     mDebugWriter->writeDebugGroup(combiner, "combiner");
-    std::cout << "Debug modules written. Press 'q' <ENTER> to stop debugging.\n" <<
+    LOG_DEBUG << "Debug modules written. Press 'q' <ENTER> to stop debugging.\n" <<
         "Press <ENTER> to write the next set...\n";
```
serin-delaunay: This (and probably the other messages in this file) should be `LOG_INFO` or whatever the user debug writer is called.

Wangscape/logging/Logging.cpp
```diff
@@ -0,0 +1,66 @@
...
+        case AppenderType::StdOut:
+        {
+            currentLogger.setStream(&std::cout);
+            break;
+        }
+        case AppenderType::StdErr:
+        {
+            currentLogger.setStream(&std::cerr);
+            break;
+        }
```
serin-delaunay: Could you add an option that appends to both `std::cout` and a file?

Wangscape/logging/Logging.cpp
```diff
@@ -0,0 +1,66 @@
...
+        case AppenderType::File:
+        {
+            if (out.is_open()) {
+                out.close();
+            }
+            out.open(name, std::ios_base::app);
```
serin-delaunay: `name` -> `filename`

Wangscape/logging/Logger.h
```diff
@@ -0,0 +1,36 @@
...
+    template<typename T>
+    Logger& operator<<(T&& msg)
+    {
+        if (mAppender != nullptr) {
+            *mAppender << std::forward<T>(msg);
+        }
+        return *this;
+    }
+private:
+    std::ostream* mAppender = nullptr;
```
serin-delaunay: How does this work when mAppender is a `std::ofstream*`? Who owns this pointer and is responsible for allocating and deallocating it? Is it possible to use a smart pointer without risking a `delete &std::cout` call?

It may be worth writing some tests to ensure that the logging system can't cause a crash.

Wangscape/logging/Level.h
```diff
@@ -0,0 +1,16 @@
...
+namespace logging
+{
+
+
+enum class Level
+{
+    Debug,
+    Info,
+    Warning,
+    Error
```
hryniuk: So... you're suggesting that `Debug` level should be for a developers and the lowest level for messages to an user should be `Info`. Do I understand it correctly? With a bit different design we could even have a different logger for our needs. But...

I'm thinking about bug reports - these "developer debugging messages" would be really helpful I think. What about this? Do you agree?

Wangscape/logging/Logging.cpp
hryniuk: Yes, of course.

Wangscape/logging/Logger.h
hryniuk: It's just a non-owning pointer to an external (global) `std::ofstream`. It's not elegant, I know, but quite convenient. It saves from having additional classes hierarchy just for appenders and allows us to have simple generic `operator<<` implementation.

Wangscape/logging/Logging.cpp
hryniuk: I'd rather want to use it as logger name, which is at the same time name of a file it's writing to. I see now, how bad designed it is :wink:. Overload will work a bit better here, I think.

Wangscape/logging/Level.h
serin-delaunay: I agree, "developer debugging messages" should be available to users so they can make bug reports. But they will be irrelevant noise when they are not encountering a bug in Wangscape, so they should be optional. That's why I think the log level should be configurable in the command line or options file.

Wangscape/logging/Logging.h
hryniuk: > Could these macros be replaced with functions or static (const?) instances/references?

Yes, of course. With my approach I've also assumed that we won't need the second logger. There are a few ways to do logging, as you most likely know, and beside static functions we could have also e.g. an `Logger` class instance and proper methods to use (like in Python: https://docs.python.org/3/library/logging.html). What do you think about that?

We could have:
* defines :x:
```cpp
LOG_DEBUG << "error no" << 1;
```
* static functions like in Qt
```cpp
logDebug() << "error no" << 2;
```
* static functions, but accepting `std::string`/`const char*`:
```cpp
logDebug("error no " + std::to_string(3));
```
* logger methods (either returning stream with `operator<<` or accepting `std::string`):
```cpp
auto logger = getLogger();
logger.logDebug("error no " + std::to_string(4));
```

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
hryniuk: Yes, that's the reason and this change of a logging level is triggered by configuration option `"DebugOutput": true`, which is connected with debug tilesets. I'm not sure, what would be proper solution - we could treat that option as one that enables "developer's logging level" or change these lines below level to `Info` and treat it as user's debug info.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
hryniuk: Yes, I agree, as I'am writing above.

Wangscape/logging/Logging.h
serin-delaunay: I prefer the Qt-style notation with static functions.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
serin-delaunay: The `DebugOutput` configuration option enables writing noise module data to the filesystem, which is slow and requires user input on `stdin`. It should be a separate option from the logging level, which may need to be set to "dev debug" due to some issue which is unrelated to noise modules.

serin-delaunay:
Great :) There are still some changes I'd like to make, but I'll open separate issues/PRs for them. What's important is that the logging system is in place :tada:

hryniuk:
Yes, finally I did it :wink:.

PR #151, opened 2017-06-11
Add command specific to OS X El Capitan

hryniuk:


README.md
```diff
@@ -174,6 +174,14 @@ First, install necessary packages with `Homebrew`:
 brew install boost sfml cmake
 ```
 
+#### Note
+On **OS X El Capitan** it may be necessary to install additional
+packages by executing ([see this question on Stack Overflow](https://stackoverflow.com/questions/32898887/boost-no-longer-works-with-homebrew-on-mac-el-capitan/32929012)):
```
serin-delaunay: I'm guessing here because I haven't had to tun this command, but I think we should name the specific package required (Xcode Command Line tools) in case the link dies.

hryniuk:
@serin-delaunay fixed.

PR #152, opened 2017-06-11
Full debug tilesets

### LINKED ISSUE #39: Output full debug tilesets

serin-delaunay:
Resolves #39.
Changes:
* Debug output is non-interactive and makes a full debug tileset for all modules in all corners.
* `example2` produces debug tilesets.
* Documented in `algorithm.md`. Should be moved to a tutorial document when we have one.

Issues:
1. Debug tilesets have a lot of repetition and redundant information, and there is no way to exclude a module or module group from debug output or mark it to output a single tile rather than a tileset.
1. Debug output is significantly slower than normal execution (but that's to be expected).
1. Debug tilesets are built simultaneously, which requires a lot of graphical memory - so much that Wangscape crashes on my laptop when trying to run `example2` with debug output (it's fine in CI). A workaround is to force main memory to be used instead, by using Mesa. Memory usage reaches over 500MB before crashing, even with 8x8 tiles and two terrains; this may be a problem with minimum texture size.
1. Debug tilesets have the same resolution as output tilesets.
1. **~Appveyor does not copy the debug output to the `example_output.zip` artefact.~**
1. **~Debug output is not mentioned in documentation.~**
1. Another callback is introduced to build debug tilesets; perhaps tileset generation should be refactored soon.
1. Output directory creation is now done in the callback defined in `Main.cpp`. ~It might be possible to remove that functionality from `OptionsManager` now.~ (That causes `TestOptions.TestOptionsValues` to fail)

I think ~5~ and ~6~ should be fixed before merging. 1, 4, and 7 can be part of a separate PR.


Wangscape/tilegen/DebugTilesetID.h
```diff
@@ -0,0 +1,61 @@
...
+        return result;
+    }
+    bool operator<(const DebugTilesetID& other) const
+    {
+        return std::tie(moduleGroupRole, moduleID, top, left) <
+            std::tie(other.moduleGroupRole, other.moduleID, other.top, other.left);
+    }
+};
+
+} // namespace tilegen
```
hryniuk: Why there's now `.cpp` file? You could hide there `ModuleGroupRole -> std::string` and `(top, left) -> std::string` mappings. I think there should be additional enum for position/orientation (what top and left are for) and then `description()` will just do

```cpp
return moduleGroupRoleComponent[moduleGroupRole] + positionComponent[position];
```
instead of these long condition checking.

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -13,8 +13,18 @@ namespace tilegen
...
+    mDebugTileBuilder(mDebugTile, {0., 0., 1., 1.})
+    // TODO (Serin) this sf::Rect is repeated in TilePartitionerNoise, factor it out
 {
+    if (options.debugOutput)
+    {
+        mDebugTile.create(options.tileFormat.resolution.x, options.tileFormat.resolution.y);
+        partition::TilePartitionerBase::DebugModuleWriter bound_debug_tile_writer = 
+            std::bind(&TilesetGenerator::writeDebugTile, this,
+                      std::placeholders::_1, std::placeholders::_2);
+        mTilePartitioner->setDebugModuleWriter(bound_debug_tile_writer);
```
hryniuk: It doesn't seem to be a good idea. You're passing a pointer to this class' private function (which modifies `mDebugTilesets`) to an another class - it have to be fixed. As I suggested below, there could be a class, which generates debug tilesets and another one, which is responsible for writing tilesets to a given directory (probably it could also write tiles from `TilesetGenerator`.

Wangscape/Main.cpp
```diff
@@ -5,6 +5,7 @@
 #include <boost/program_options/options_description.hpp>
 #include <boost/program_options/parsers.hpp>
 #include <boost/program_options/variables_map.hpp>
+#include <boost/filesystem.hpp>
```
serin-delaunay: Can clang-format check include order automatically?

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -13,8 +13,18 @@ namespace tilegen
 TilesetGenerator::TilesetGenerator(const Options& options,
                                    std::unique_ptr<partition::TilePartitionerBase> tile_partitioner) :
     options(options),
-    mTilePartitioner(std::move(tile_partitioner))
+    mTilePartitioner(std::move(tile_partitioner)),
+    mDebugTileBuilder(mDebugTile, {0., 0., 1., 1.})
+    // TODO (Serin) this sf::Rect is repeated in TilePartitionerNoise, factor it out
```
serin-delaunay: Here: https://github.com/Wangscape/Wangscape/blob/master/Wangscape/tilegen/partition/TilePartitionerNoise.cpp#L141. It defines the region where noise values are evaluated.

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -30,21 +40,31 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
...
         // relative to output dir, not options dir!
-        metaOutput.addTileset(clique, filename, res_x, res_y);
+        metaOutput.addTileset(clique, filename, mResX, mResY);
         generateClique(clique, *output, filename);
         output->display();
 
         p.append(filename);
         callback(output->getTexture(), p.string());
         p.remove_filename();
+        for (auto& it : mDebugTilesets)
```
serin-delaunay: I don't think that would work, because the non-`const` method `display()` is called on `*it.second`.

Wangscape/tilegen/partition/TilePartitionerNoise.cpp
```diff
@@ -109,24 +106,13 @@ void TilePartitionerNoise::noiseToAlpha(std::vector<noise::RasterValues<double>>
...
-    mDebugWriter->writeDebugGroup(central, "central");
-    mDebugWriter->writeDebugGroup(border_h, "border_h");
-    mDebugWriter->writeDebugGroup(border_v, "border_v");
-    mDebugWriter->writeDebugGroup(combiner, "combiner");
-    logInfo() << "Debug modules written. Press 'q' <ENTER> to stop debugging.\n" <<
-        "Press <ENTER> to write the next set...\n";
-    int keypress = std::cin.get();
-    if (keypress == 'q')
+    if (!mDebugModuleWriter)
+        logError() << "Requested debug data but didn't provide a debug module writer function!\n";
```
serin-delaunay: The refactor you suggest above will probably make this check unnecessary.

serin-delaunay:
I agree that after these changes the tileset generation code needs to be refactored.

Currently it works like this:
* `TilesetGenerator` stores main tileset and debug tilesets.
* `TilesetGenerator` loops over cliques and corner configurations:
  * `TilesetGenerator` calls `TileGenerator`, which calls `PartitionerNoise`. Return value is used to build a tile.
  * `PartitionerNoise` calls a `TilesetGenerator` function from inside `PartitionerNoise::makeCornerModule` to build debug tiles. This is the only context where individual noise modules and their names are accessible. The `TilesetGenerator` function is passed to `PartitionerNoise` in order to avoid creating a circular dependency between the two classes.

I think it would make sense to split `TilesetGenerator` into two parts. The first (`TilesetGenerator`?) would drive the loop which calls `PartitionerNoise`, and the other (`TilesetImages`?) would contain the tilesets, debug tilesets, and methods to add tiles to them (possibly with the help of `TileGenerator`). Both `TilesetGenerator` and `PartitionerNoise` could have access to `TilesetImages`.

Saving tilesets to file could be done in `TilesetImages`. I don't see why it would need a separate class.

Wangscape/Main.cpp
hryniuk: Yes, it can even reorder them, which can be a git hook.

Wangscape/tilegen/TilesetGenerator.cpp
hryniuk: Oh, ok. This comment is placed below line it describes, not above as usual and there's `{}` initialization, so I haven't noticed it and I was wondering which variable stores that Rect 😄.

Wangscape/tilegen/TilesetGenerator.cpp
hryniuk: It compiles and seems to work for me. The thing is, if I reason correctly, with `const auto&` you'll have a const reference to the `std::pair<DebugTilesetID, std::unique_ptr<sf::RenderTexture>>` and while you **cannot** call any non const method that will change value of this **unique ptr**, methods of underlying object don't change a value of it, so it's not a problem. Have you tried it on you compiler?

---

Example
```cpp
#include <iostream>
#include <memory>

struct A
{
    void fx() {++x_;}
    int x_ = 0;
};

struct B
{
    B() : a{std::make_unique<A>()}
    {}
    std::unique_ptr<A> a;
    int y_ = 0;
};

int main()
{
    const B b1;
    const B& b2 = b1;
    std::cout << b2.y_ << std::endl;
    b2.a->fx();
    std::cout << b2.a->x_ << std::endl;
}
```

On `gcc 7.1.1` and `clang 4.0.1` this snippet produces:

```shell
0
1
```

Wangscape/tilegen/TilesetGenerator.cpp
```diff
@@ -30,21 +40,31 @@ void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::stri
...
+        {
+            it.second->display();
+            std::stringstream debug_filename_ss;
+            debug_filename_ss << filename_base << it.first.description() << "." << options.tileFormat.fileType;
+            auto debug_file_path = (p / "debug" / debug_filename_ss.str()).string();
```
hryniuk: It could be also `const` I think :wink:.

hryniuk:
Ok, I think we can discuss it after merging this PR in some issue. I see it almost in the same way as you. The main difference I think is that in my opinion regular and debug tilesets should be splitted. They share the same algorithm of generating them, but they have different purposes. For me there should be:

* one class for a `Tile` and another class that stores `Tileset` (no matter how it's produces, if it's regular or debug tileset)
* one class for generating `Tileset`, let's say `TilesetGenerator`, which gets a strategy of creating them. So current `void TilesetGenerator::generate(std::function<void(const sf::Texture&, std::string)> callback)
` will become more like `Tileset TilesetGenerator::generate(std::function<void(const sf::Texture&, std::string)> callback)`. It could get a list of ways/strategies/algorithms of generating tilesets and return a list of `Tileset`s, one for each strategy. I'm not sure about this, but it seems to be the most clear solution for me.
* one class for saving tilesets to a given file

> Saving tilesets to file could be done in TilesetImages. I don't see why it would need a separate class.

In my opinion it should be a separate class, cause:
* just name `TilesetImages` suggests that it shouldn't do anything with filesystem
* not all `TilesetImages` will be saved to a disk - for example with GUI we can just display them for a preview
* with a separate class we [won't need to change `TilesetImages` class, when we will have to change the way they're saved](https://en.wikipedia.org/wiki/Single_responsibility_principle) (we could want to compress it or change output format in the future, depending on user preferences)
* it will make `TilesetImages` smaller and easier to test and maintain. Moreover by keeping one responsibility for each class, different contributors will be able to work on a different layers of a project (e.g. input/output and tile generation) without problems with changing the same class.

serin-delaunay:
I think that's all the issues taken care of, apart from the `TilesetGenerator` refactoring. We can open an issue to work out the details of that.

Wangscape/tilegen/DebugTilesetID.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace tilegen
+{
+
+struct DebugTilesetID
+{
+    ModuleGroupRole moduleGroupRole;
+    noise::ModuleID moduleID;
+    bool top;
+    bool left;
```
hryniuk: Why do you want to keep them public?

serin-delaunay: They're used as `std::map` keys, so those values shouldn't be writeable. I've made them private.

PR #158, opened 2017-07-13
New modules

### LINKED ISSUE #88: Add more modules
- [x] Divide: two source modules, outputs the first divided by the second. Option to perform zero divisions silently?
- [x] Reciprocal: one source module, outputs one divided by the source value. Option to perform zero divisions silently?
- [x] Logarithm: one source module, outputs base-b logarithm of the source value. Parameter to control the base. Option to output NaN rather than throw an exception when value or base is non-positive?

serin-delaunay:
Resolves #88. Adds a `Bitmap` module which will be useful on its own, but is also required for #32.

TODO:
- [x] Add tests.:
  - [x] Logarithm, Reciprocal, Divide
  - [x] Bitmap
- [x] Give `Bitmap` access to `Options` so that it can retrieve image files from the correct folder. Maybe `OptionsManager` or `Options` should ~be a singleton~ have a global instance?
- [x] Store `Bitmap` data in a greyscale structure ~extracted from `RasterValues` into a separate class~ shared with `RasterValues`.
- [x] Give `Bitmap` options to scale the output values by the maximum value in image.
- [x] Integrate `Bitmap` with module decoding classes (as described in #88).
- [x] Deal with infinity and NaN reasonably in alpha calculators.
- [x] Add `image/` folder to CMake
- [x] Add Armadillo to CMake
- [x] Install Armadillo on Travis:
  - [x] MacOS
  - [x] Ubuntu


Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+    return mFilename;
+}
+
+void Bitmap::SetFilename(const std::string & filename)
+{
+    // Load and process file if not already done
+    auto it = loadedImages.find(filename);
+    if (it == loadedImages.cend())
+        return;
+    else
```
hryniuk: I'd remove this `else` clause as you're returning early above. One thing is that `else` is not needed and the second is that `if` part is only preconditions/error checking and the second, `else` part is the "real" part of this function.

serin-delaunay: There are cases where an unnecessary `else` is clearer, but you're right that this is probably not one of then.

Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        auto bitmap = std::make_shared<sf::Image>();
+        // TODO get base file path from Options. How?
+        bitmap->loadFromFile(filename);
+        for (size_t y = 0; y < bitmap->getSize().y; y++)
```
hryniuk: I see that's it's the same problem I've faced in my [image-processing](https://github.com/hryniuk/image-processing) project - looping through an image is repeated everywhere and there are a few special operations that you'd like to do on it... I've used instances of `QImage` class there and I've created functions accepting them to do e.g. looping through the whole image with a function (something like `std::for_each`).

In Wangscape, I think, we could have an abstraction over image that will hide the exact type of it and will provide some useful methods as `isGreyscale()`, which could be used in this line. What do you think about it?

Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+    mDefaultValue = defaultValue;
+}
+
+double Bitmap::GetValue(double x, double y, double z) const
+{
+    if (mRegion.contains(x, y))
+        return mDefaultValue;
+    if (mResolution.x == 0 || mResolution.y == 0)
+    {
+        logWarning() << "Tried to call Bitmap::GetValue with zero-pixel resolution";
```
hryniuk: I wonder if it's not a good place to `throw`, cause for me it's something like calculating a square of a negative in real numbers.

Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+{
+    return mDefaultValue;
+}
+
+void Bitmap::SetDefaultValue(double defaultValue)
+{
+    mDefaultValue = defaultValue;
+}
+
+double Bitmap::GetValue(double x, double y, double z) const
```
hryniuk: `z` is not used, if I see correctly.

serin-delaunay: Yes. But the function must conform to `Module::GetValue`.

Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    offset.x *= mResolution.x;
+    offset.y *= mResolution.y;
+
+    return getPixel(static_cast<size_t>(offset.x), static_cast<size_t>(offset.y));
+}
+
+double Bitmap::getPixel(size_t x, size_t y) const
+{
+    if (!mImage)
```
hryniuk: Is this `if` needed? I mean it's `Bitmap` class, so as an user I would expect that if I have an instance of this class, it's valid and has an image assigned to it. I'd rather set `mImage` in constructor and don't provide the way to set it or make every reading operation explicit by throwing an exception on fail to keep `mImage` valid.

Wangscape/noise/module/Bitmap.cpp
serin-delaunay: Yes, failing this check would be a clear configuration error and should crash. However, I would hope that a module configuration which calculates `(-1)^(1/2)` would return `NaN` (and I might add a test to check that on the Wangscape `Pow` module). I think this is more like `Clamp` with incompatible min and max values (which IIRC throws an error or crashes).

Wangscape/noise/module/Bitmap.h
```diff
@@ -0,0 +1,47 @@
...
+    void SetRegion(sf::Rect<double> region);
+
+    double GetDefaultValue() const;
+    void SetDefaultValue(double defaultValue);
+
+
+    virtual double GetValue(double x, double y, double z) const;
+private:
+    double getPixel(size_t x, size_t y) const;
+    std::string  mFilename;
```
hryniuk: This `mFilename` bothers me. Why is it needed? What's a purpose of saving it? By looking at an interface of this class it's hard for me to guess that this class beside storing an image also loads it and it does so in the `SetFilename()` function - it seems to be only a setter of `mFilename` variable and I'd suppose that there would be also a `SetImage` function.

You're caching these images in `loadedImages`, so maybe there's no need to store `mImage` in an instance of this class (which of course would be weird in **Bitmap** :wink:), only a filename - key of that map.

Wangscape/noise/module/Bitmap.cpp
hryniuk: Yes 😅. When there are two decisions "on the same level", prime/composite number for example.

I used to implement a "main body" of a function under `if`, which checked a positive variant, leaving error handling for the rest of a function, but next I've switch to checking preconditions at the beginning as you've done here and that's a better way I think.

Wangscape/noise/module/Bitmap.h
serin-delaunay: The filename should be stored in order to conform with the `noise::Module` interface (where each attribute has a `Get` and `Set` method).

It's not strictly necessary, but `GetValue` is called in inner loops (typically once per pixel per module), so I would prefer to replace the `map::find` call with a simple pointer dereference - even if the planned use cases only have a few items in the `map`.

Wangscape/noise/module/Bitmap.cpp
serin-delaunay: `Bitmap`'s image processing is pretty lightweight, but it's part of the solution to #32, which will use a lot of operations like this. The [draft](https://gist.github.com/serin-delaunay/a04417c5bd2ef9dad64d01f867c2ee19) uses `numpy.ndarray` with `skimage` to abstract all of these loops away, and I would hope to do something similar in the C++ implementation. I'd prefer to avoid bringing the whole of Qt into Wangscape, but a tensor/matrix/linalg library like Eigen, Armadillo, CImg, or even Boost.multi_array would probably be ideal.

Wangscape/noise/module/Bitmap.cpp
serin-delaunay: The `noise::Module` interface requires trivial constructors, with attributes assigned by `Set` methods afterwards. The constructor could only set `mImage` to a null image (probably 1x1 pixel with value 0). There are plenty of other `noise::Module` classes which are invalid with default attributes (`Curve`, for example).

serin-delaunay: Also, the tested condition should be (`mImage`).

Wangscape/noise/module/Bitmap.cpp
serin-delaunay: Chose `Armadillo` because it has third-order tensors. `xtensor` produces security warnings in copy operations, `DyND` doesn't compile, `CImg` has a weird license, `Eigen`'s tensor support isn't mature, `Boost.multi_array` doesn't have much support for vectorised operations.

serin-delaunay:
I think this is ready for review. It works on OSX Travis, so the Ubuntu Travis build probably just needs https://github.com/Wangscape/wangscape-build-dockerfile/pull/3.

I've replaced the `OptionsManager` allocated on the stack in `main()` with a global instance. It's not a singleton, so additional instances can be constructed if required in future. All the objects with a `const Options&` assigned in the constructor can now access it through `getOptionsManager().getOptions()` instead. That would be quite a far-reaching change, so it might be better in a separate PR.

Wangscape/CMakeLists.txt
```diff
@@ -33,3 +35,9 @@ if(SFML_FOUND)
     include_directories(${SFML_INCLUDE_DIR})
     target_link_libraries(Wangscape ${SFML_LIBRARIES})
 endif()
+
+find_package(Armadillo REQUIRED)
+if(SFML_FOUND)
```
hryniuk: Should be `Armadillo_FOUND` I suppose :wink:.

serin-delaunay: All caps according to https://github.com/Kitware/CMake/blob/master/Modules/FindArmadillo.cmake

Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,145 @@
...
+    mMaxScale = max_scale;
+    if (mImage && max_scale)
+        updateMax();
+}
+
+double Bitmap::GetValue(double x, double y, double z) const
+{
+    for (auto v : *mImage)
+    {
+        std::cout << int(v) << " ";
```
serin-delaunay: That line can just be removed.

Wangscape/noise/module/Bitmap.cpp
```diff
@@ -0,0 +1,145 @@
...
+    sf::Vector2<double> offset = {x, y};
+
+    offset.x -= mRegion.left;
+    offset.y -= mRegion.top;
+
+    offset.x /= mRegion.width;
+    offset.y /= mRegion.height;
+
+    offset.x *= mImage->n_cols;
+    offset.y *= mImage->n_rows;
```
serin-delaunay: It would also be nice to vectorise these operations, but that's part of #46.

Issue #17, opened 2016-12-10
GUI interface for option specification


serin-delaunay:
At the moment all configuration is done through editing a JSON file. This tool should be usable by non-programmers, so a GUI interface will eventually be needed.

Ideally, such a tool should also provide quick feedback about the effects of altered options - showing one or more example tiles immediately rather than writing all the tileset images.

It may also be nice to have the tileset example application (#10) accessible from or integrated with the GUI editor. But it would be OK to wait for full tileset generation before using that.

- [ ] GUI option editor
- [ ] Integrated example tile generator
- [ ] Access/integrate with landscape viewer (#10)

serin-delaunay:
This will be low-priority until the options specification becomes less volatile, or a solution is found that autogenerates the GUI from the options schema.

hryniuk:
How do you feel about using **Qt** for this? It will be a nice learning opportunity for me 😄. We could discuss input options during this week and I think I'll be able to start around the coming weekend.

serin-delaunay:
Qt would be a good choice for this.

serin-delaunay:
Closed in #159. There are still points here to resolve, but they can be split into separate issues now.

Issue #150, opened 2017-06-09
Make AppVeyor builds report failure if examples fail Wangcheck


serin-delaunay:
The issue resolved by #149 causes one of the examples to fail Wangcheck (https://ci.appveyor.com/project/serin-delaunay/wangscape/build/0.1.194#L2161), but AppVeyor reports that the build succeeded. AppVeyor should report failure in such cases.

PR #159, opened 2017-07-19
Add QT GUI

hryniuk:
Includes #17, eventually providing a tileset preview and options editor in a window.


hryniuk:
Things to do:
- [x] write `CMakeLists.txt` to include QT libraries into building process
- [x] add a button for tileset generation
- [ ] add a preview area for displaying generated tileset
- [ ] add options editor (at least some basics):
    - [ ] add a selector for options directory 
    - [ ] add a selector for output directory
- [ ] disable GUI building in CI or install and set up there QT libraries

hryniuk:
@serin-delaunay could you for now only check, whether this branch compiles and shows a window on your machine? Path to options directory is temporarily hard-coded, so you have to change it in code and recompile to generate a tileset using button.

Note that in `CMakeLists.txt` I've just included `wangscape-main` library. I wasn't sure, whether it's a good idea to split Wangscape into small libraries, one for each directory at that time; now we can just include different `main.cpp` file and create a GUI using this that lib :smiley:.

WangscapeGui/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+qt5_wrap_ui(UIS_HDRS mainwindow.ui)
+
+# Tell CMake to create the helloworld executable
+add_executable(WangscapeGui WIN32 main.cpp mainwindow.cpp ${UIS_HDRS})
+
+# Use the Widgets module from Qt 5.
+target_link_libraries(WangscapeGui wangscape-main)
+
+target_link_libraries(WangscapeGui Qt5::Widgets)
```
hryniuk: It should try to find a `Qt5` library, instead of assuming that it's installed.

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <iostream>
+
+#include <iostream>
+#include <fstream>
+#include <memory>
+
+#include <boost/program_options/options_description.hpp>
+#include <boost/program_options/parsers.hpp>
+#include <boost/program_options/variables_map.hpp>
```
hryniuk: It's not needed, cause it doesn't use `program_options`.

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    ui->setupUi(this);
+    connect(ui->pushButton, SIGNAL(pressed()), this, SLOT(clickPushButton()));
+}
+
+void MainWindow::clickPushButton()
+{
+    logging::addAppender(std::make_unique<logging::ConsoleAppender>("console", logging::Level::Debug));
+
+    std::string filename = "/home/hryniuk/projekty/Wangscape/doc/examples/example2/example_options.json";
```
hryniuk: It have to be read form a text field.

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,66 @@
...
+    QMainWindow(parent),
+    ui(new Ui::MainWindow)
+{
+    ui->setupUi(this);
+    connect(ui->pushButton, SIGNAL(pressed()), this, SLOT(clickPushButton()));
+}
+
+void MainWindow::clickPushButton()
+{
+    logging::addAppender(std::make_unique<logging::ConsoleAppender>("console", logging::Level::Debug));
```
hryniuk: Do we need to use a logger? We could think about notifications using Qt windows.

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    OptionsManager optionsManager(filename);
+    const Options& options = optionsManager.getOptions();
+
+    std::unique_ptr<tilegen::partition::TilePartitionerBase> tp =
+        std::make_unique<tilegen::partition::TilePartitionerNoise>(options);
+
+    tilegen::TilesetGenerator tg(options, std::move(tp));
+
+    std::cout << "generating tiles..." << std::endl;
```
hryniuk: Remove this debug print.

WangscapeGui/mainwindow.cpp
serin-delaunay: I think while it's in the experimental stage we should use a logger. Log output should be viewable through something like a log widget or nonmodal window before public release.

WangscapeGui/mainwindow.cpp
serin-delaunay: Can we get the filename from a standard file selector dialogue?

hryniuk: Yes, of course, I've only meant that it will be read from GUI.

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,66 @@
+#include "mainwindow.h"
+#include "ui_mainwindow.h"
+
+#include <iostream>
+
+#include <iostream>
```
serin-delaunay: The second `#include <iostream>` can be removed.

hryniuk:
Screenshot :).

tmp sljy1xveoi

serin-delaunay:
I made a project file for d2794d84d4d01e97da783723e51327685d487e4a, and it compiles and writes tilesets. However, it does use individual invocations of `uic` and `moc` for each file, which might not scale well. Screenshot: 
wangscape-gui

serin-delaunay:
(Also, this is awesome!)

hryniuk:
Build on Linux finished successfully, so I'm removing "(WIP)" prefix. Two things left: green osx build and small refactoring (probably).

hryniuk:
I've just noticed that `Options` can be easily copied and `MainWindow` can work on the copy by recreating `TilePartitioner` and `TilesetGenerator` after each change of e.g. alpha options. One of the problems I've seen was that `MetaOutput::createOutputDirectory` was called with options loaded from a file, but the directory can be created independently. I'll add an editor of options I've mentioned above to the GUI.

hryniuk:
I've find out, which flag I'm supposed to pass to `cmake` to build Qt 5 Widgets project 😃, but there's another problem as I've mentioned in checkbox - linking failure make build green anyway, see e.g. https://travis-ci.org/hryniuk/Wangscape/jobs/266115590.

serin-delaunay:
We have a similar problem on Appveyor (#150). I think it's to do with how CI interprets program return values, see eg. https://github.com/travis-ci/travis-ci/issues/6768 .

hryniuk:
@serin-delaunay yes, it should be easy to fix on Linux and OS X - just by saving return value and adding one `if`.

hryniuk:
I've moved buttons to menu, it looks better. I plan to add another action in menu, opening a window with options editor, one per tab. First tab: **Alpha Calculator Mode**.

hryniuk:
@serin-delaunay Ok, it's ready :d. I don't want to lengthen the time to merge. There are some issues, that in my opinion can be resolved in other PRs, like zooming preview area or refactoring progress bar (to use it for tileset generation not only conversion of preview images and set scale adequate to images' sizes). That build's return code, as it's shared between different build environments, can be fixed later too, I think.

Almost 800 lines in one go is too much anyway. There's alpha mode editor and default directory flag (it asks the user for output directory, when it's unchecked) and that's enough for now ;). I'm trying to build it in QtCreator on Windows now.

README.md
```diff
@@ -95,6 +95,8 @@ Wangscape depends on these libraries:
   - graphics
 * **Armadillo**
   - **OpenBLAS**
+* **Qt5**
+>>>>>>> Add Qt5 to OSX build and README
```
serin-delaunay: Can this line be removed?

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,202 @@
...
+{
+    delete mUi;
+}
+
+void MainWindow::generate()
+{
+    if (mTilesetGenerator == nullptr)
+    {
+        QMessageBox error_message;
+        error_message.critical(this, "Error during tileset generation", QString("You have to select options file before clicking 'Generate...' button"));
```
serin-delaunay: "You must select an options file before clicking the 'Generate...' button"

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,202 @@
...
+    resetTilesetGenerator();
+
+    mUi->comboBox->clear();
+    mPreviewImages.clear();
+
+    mTilesetGenerator->generate([this](const sf::Texture& output, std::string filename)
+    {
+        const sf::Image source_image = output.copyToImage();
+
+        const auto basename = filename.substr(filename.rfind('/') + 1);
```
serin-delaunay: Can we use `boost::filesystem::path` here? This will break if `\` is used as a path separator.

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+void MainWindow::saveOutput()
+{
+    if (mOptionsEditor->useDefaultOuputDir())
+    {
+        mOptions.relativeOutputDirectory = mOriginalOptions.relativeOutputDirectory;
+    }
+    else
+    {
+        QString chosen_path = QFileDialog::getExistingDirectory(this, tr("Select output dir"));
```
serin-delaunay: Unless there are space constraints, user-facing text should say "directory".

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,202 @@
...
+        mOptions.relativeOutputDirectory = chosen_path.toLocal8Bit().constData();
+    }
+
+    if (mPreviewImages.empty())
+    {
+        QMessageBox warning_message;
+        warning_message.warning(this, "Warning", QString("You have to generate tileset before saving it"));
```
serin-delaunay: "You must generate a tileset before saving it"

WangscapeGui/mainwindow.cpp
```diff
@@ -0,0 +1,202 @@
...
+    QString chosen_path = QFileDialog::getOpenFileName(this,
+                                                     tr("Select options file"));
+    if (chosen_path.isEmpty())
+        return;
+
+    mUi->currentOptionsPath->setText(QString(chosen_path));
+
+    // TODO(hryniuk): make OptionsManager throw custom exception, catch it here
+    // and show QMessageBox with a proper error message
+    mOptionsManager = std::make_unique<OptionsManager>(chosen_path.toLocal8Bit().constData());
```
serin-delaunay: We have a global `OptionsManager` instance now; is it possible to use that instead?

WangscapeGui/mainwindow.h
```diff
@@ -0,0 +1,52 @@
...
+
+    Ui::MainWindow *mUi;
+    QGraphicsScene *mScene;
+    // TODO(hryniuk): create a struct for this map's elements
+    std::map<std::string, std::pair<std::string, QImage>> mPreviewImages;
+    QString mOptionsFilePath;
+    std::unique_ptr<OptionsManager> mOptionsManager;
+    Options mOriginalOptions;
+    Options mOptions;
+    OptionsEditor* mOptionsEditor;
```
serin-delaunay: Is it possible to use smart pointers for `mUi`, `mScene`, and `mOptionsEditor`?

WangscapeGui/optionseditor.cpp
```diff
@@ -0,0 +1,64 @@
+#include "optionseditor.h"
```
serin-delaunay: Is all-lowercase for window classes a Qt convention, or can we rename these to `OptionsEditor.h`, `MainWindow.cpp`, etc.?

WangscapeGui/optionseditor.h
```diff
@@ -0,0 +1,33 @@
...
+    void setOptions(Options* options);
+
+private slots:
+    void setAlphaCalculatorMode(const QString &mode_name);
+    void switchDefaultOutputDirFlag(int newCheckboxState);
+
+private:
+    void setAlphaCalculatorModeComboBox();
+    Ui::OptionsEditor *mUi;
+    Options* mOptions;
```
serin-delaunay: Smart pointers please.

README.md
hryniuk: Oh, of course.

WangscapeGui/optionseditor.cpp
hryniuk: They were generated by QtCreator, I'll change them.

WangscapeGui/mainwindow.h
hryniuk: As far as I know yes, but it's **not recommended**, cause Qt has it's own memory management system:
https://stackoverflow.com/questions/34433435/why-dont-the-official-qt-examples-and-tutorials-use-smart-pointers

I'm not really sure about mUi (from what I've read I conclude that it's handled by Qt too), but others are QObjects, thus they are properly released.

WangscapeGui/optionseditor.h
hryniuk: Yes, you're right here. I thought that it's not owning pointer and this class is managed by it's parent, `MainWindow`, so it will be ok, but:
* it may change in the future and
* storing it in `std::shared_ptr` in this place is just a bit safer :smiley:.

WangscapeGui/optionseditor.cpp
hryniuk: Auto-generated headers file, `ui_*`, will be still all lower-case.

hryniuk: Ops, wrong. They were just some cached version, fixed to `ui_MainWindow.h` and `ui_OptionsEditor.h`.

hryniuk:
There's a problem with OSX build connected to stdlib. I thought that it should be fixed with `-DCMAKE_CXX_FLAGS="-stdlib=libc++"` flag, but it failed. I'll try to find a solution tomorrow.

See:
https://stackoverflow.com/questions/10991274/initializer-list-in-clang
https://stackoverflow.com/questions/14790354/is-initializer-list-like-this-legal-in-c11
and, of course, Travis build:
https://travis-ci.org/Wangscape/Wangscape/jobs/268994710

hryniuk:
I've just move constructing the map to a function, resigning of initializer_list for now.

Wangscape/tilegen/alpha/CalculatorMode.h
```diff
@@ -13,5 +15,7 @@ enum class CalculatorMode
     Dither
 };
 
+extern const std::map<std::string, CalculatorMode> calculatorModeByName;
```
serin-delaunay: You need `#include <string>`. Clang's error messages for this are terribly unhelpful :(

hryniuk: Eh... I was misled by a fact it builds on `gcc`. Headers seemed to be OK.

WangscapeGui/MainWindow.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }());
+}
+
+QImage MainWindow::convertSfImageToQImage(const sf::Image& source_image)
+{
+
+    const sf::Vector2u output_image_size = source_image.getSize();
+    QImage output_image(output_image_size.x, output_image_size.y, QImage::Format_RGB32);
+
+    for (int y = 0; y < output_image_size.y; ++y)
```
serin-delaunay: `unsigned int y` to match `output_image_size.y`; same for `x`.

hryniuk: I think we should build it with `-Wall` enabled and probably, when all warnings will be fixed, with `-Werror`.

serin-delaunay: We used to have a warning from `googletest` during `WangscapeTest` compilation, but in the latest `master` build that error seems to be gone :)

serin-delaunay:
I've adapted the VS project to the new changes, and it builds fine. A few comments on the interface (probably they should be separate issues):
1. "Select options file" is hidden behind the "File" menu; should be an obvious button.
1. When loading the options file, a label at the bottom has text set to a file path. If the path is long, the whole window expands to allow it to be displayed. The effect is quite jarring, and very long paths could cause problems, so I think the path should be cut off, or the label should have multiple lines.
1. The program becomes unresponsive during tileset generation; could that be done in a separate thread?
1. `alphaCalculatorTopTwoPower` should be in the `Alpha` tab of the options editor when `TopTwo` or `Dither` is selected.
1. With "Use default output directory" turned off, if I try to save the tileset I get a folder selector widget named "Select output directory". If I type the name of a nonexistent folder there, an error message is displayed. Instead, it should create the folder and return that path.
1. When quitting the program (after triggering the above message and saving successfully to a non-default output directory), I get a `boost::filesystem::filesystem_error`, apparently generated at `MainWindow.cpp:95` (`boost::filesystem::create_directories(image_dir);`). This was from a release build, so the error may in fact occur on a different line.
1. `CalculatorMode.cpp` should be included in the `Wangscape` VS project.
1. Finally, this is really awesome!

I think 6. and 7. should be fixed, and then I'm happy to merge this. If you want to include a QtCreator project or my VS project files (https://github.com/serin-delaunay/Wangscape/tree/gui-vs-2) for WangscapeGui, that's OK too.

hryniuk:
Thanks a lot for testing. I've included your commit and fixed 6. issue. I'll create separate issues for all points you mentioned and my suggestions I've wrote down during implementation.

serin-delaunay:
Looks like my VS project isn't working on AppVeyor. It's fixed in https://github.com/Wangscape/Wangscape/pull/172/commits/3a0ae18cf02e6084349fe7d75b309dbcebac6cce.

serin-delaunay:
Merged. Great work! :tada:

PR #163, opened 2017-08-24
New tile rearranger

serin-delaunay:
Required for #32.
* More Armadillo-based image processors and converters with tests (~`copyRegion` still needs tests~).
* `TileRearranger` computes dual partition and leaves a `TileRearrangement` which can be applied to `sf::Image`s.
* Dual tiles can have any geometry which tessellates like a parallelogram.
* Use C++11 Armadillo features.
* Enable logging in WangscapeTest (but it still doesn't output anything in VS test runs).
* Uses clockwise rather than binary corner ordering; 3- and 6-corner versions will not be compatible with binary corner codes. Main tile generation still uses binary ordering.

To do:
- [x] Test `copyRegion` and `copyRegionBounded`.
- [x] Find edges and corners of dual tiles.
- [x] Find distance fields of edges and corners:
  - [x] L_1 distance.
  - [x] L_infinity distance.
  * ~L_2 distance using fast marching method~ (separate issue).
- [x] Add more complex test cases for tile rearrangement (from https://gist.github.com/serin-delaunay/3da9c4838e9e489d7ea1c4b22373c8cc)
  - [x] Fix bug in texture rearrangement (parts of some regions are cut off).
- [x] Replace current tile rearranger with this (input would still be just rectangles).
- [x] Replace `sf::Vector` and `arma::SizeMat` use with `vmml::Vector<2,T>` in new code (fully resolving #46 can be a separate PR).
- [x] Move `tessellated()` out of `TileRearranger` into `Image`.
  - [x] Test `tessellated()`.
- [x] Hide `dilatedEroded()`.
- [x] Switch `offsetA` and `offsetB`. The first should be "horizontal".
- [x] Add comments ~or document the process in `algorithm.md`~.
- [x] Remove templating of `TileRearranger`. Making 3- and 6-corner templated versions will require pairs of objects for input/output values, and making `TileRearranger<4>` compatible with that is beyond the scope of this PR.
- [x] Make `sf::Image` comparison in tests less verbose; too many disagreements can flood the console.
- [x] Update `CMakeLists.txt`.
- [x] Include VMMLib license in binary distribution (merge conflict with #164)
- [x] Fix Armadillo errors in GCC builds (https://travis-ci.org/Wangscape/Wangscape/jobs/268387507). This will require using an up-to-date Armadillo version with `index_max` and `min(cube)`.
- [x] Update `algorithm.md` with diagrams of tile rearrangement (including motivation).
- [ ] Add more comments to rearrangement/image processing code (where required).

Implementing full support for non-rectangular 4-cornered tiles will also require (probably in a separate PR):
* User provides tile partition image and offset vectors in options file.
* `Bitmap`-related module (common base class) which holds distance from tile edges and corners.
* Replace `corner_combiner` and `edge_favouring_mask` modules in examples with distance bitmaps.
* Mask noise evaluation and output tiles using dual tile.
* Switch to clockwise corner order everywhere.
* Include offset vectors in metaoutput.
* Update Wangview:
  * Use offset vectors in display
  * Generate map large enough to fill the window (possibly parallelogram-shaped).

### DIFF
--- a/Wangscape/image/Image.cpp
+++ b/Wangscape/image/Image.cpp
@@ -18,7 +18,7 @@ std::function<bool(IVec)> boundsChecker(IVec size, IVec origin = {0, 0})
 
 ImageStackGrey generateDisplacements(const ImageGrey & image, const std::vector<IVec>& offsets, bool include_centre, arma::u8 edge_value = 0)
 {
-    unsigned int c = include_centre;
+    const unsigned int c = include_centre ? 1 : 0;
     ImageStackGrey displacements(image.n_rows, image.n_cols, offsets.size() + c);
     displacements.fill(edge_value);
     if (include_centre)
@@ -301,5 +301,5 @@ ImageGrey32 distances(const ImageGrey & traversable, const ImageGrey & sources, 
 ImageStackGrey tessellated(const ImageGrey & image, const std::vector<IVec>& offsets)
 {
     const auto image_padded = padded(image);
-    return generateDisplacements(image,offsets,false);
+    return generateDisplacements(image_padded,offsets,false);
 }


Wangscape/image/Image.cpp
```diff
@@ -30,3 +63,254 @@ ImageGrey imageToGreyscale(const ImageColour & image)
...
+    std::queue<IVec> border;
+    const auto& steps = use_diagonals ? ORTHODIAGONAL_STEPS : ORTHOGONAL_STEPS;
+    const auto in_bounds = [&shape](IVec point)
+    {
+        if (point.x() < 0 || point.y() < 0)
+            return false;
+        if ((unsigned int)point.x() >= shape.x() || (unsigned int)point.y() >= shape.y())
+            return false;
+        return true;
+    };
```
hryniuk: It should be a standalone function.

Wangscape/image/Image.cpp
```diff
@@ -30,3 +63,254 @@ ImageGrey imageToGreyscale(const ImageColour & image)
...
+                continue;
+            if (distance_field(new_point.y(), new_point.x()) == infinity)
+            {
+                border.push(new_point);
+                distance_field(new_point.y(), new_point.x()) = distance;
+            }
+        }
+    }
+    return distance_field;
+}
```
hryniuk: Add newlines to separate parts of this function. It will be much easier to read :).

Wangscape/image/Image.cpp
```diff
@@ -30,3 +63,254 @@ ImageGrey imageToGreyscale(const ImageColour & image)
...
+            if (!in_bounds(new_point))
+                continue;
+            if (!traversable(new_point.y(), new_point.x()))
+                continue;
+            if (distance_field(new_point.y(), new_point.x()) == infinity)
+            {
+                border.push(new_point);
+                distance_field(new_point.y(), new_point.x()) = distance;
+            }
+        }
```
hryniuk: I'd consider extracting this neighborhood-mapping part... but I'm not sure how (what arguments it should accept). I suppose that it will be repeated one by one in future code.

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
...
+    sf::Color::Green,
+    sf::Color::Blue,
+    sf::Color::Yellow,
+    sf::Color::Cyan,
+    sf::Color::Magenta};
+
+TileRearranger::TileRearranger()
+{
+    for (int i = 0; i < Corners; i++)
+        regionColours[i] = default_region_colours[i];
```
hryniuk: `std::copy`

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
+#include "TileRearranger.h"
+#include <set>
+#include "logging/Logging.h"
+const static std::array<sf::Color, 6> default_region_colours{
```
hryniuk: Add a newline before this line.

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
...
+    // Tessellate padded version of dual tile in 3x3 grid
+    const auto dual_tessellation = tessellated(mTileRearrangement.dual, tessellation_offsets);
+    // Check that tessellation has no holes or overlaps
+    validateTessellation(dual_tessellation);
+
+    // Use adjacency between dual tile mask and offset versions to find edges and corners of dual tile mask
+    findDualBoundaries(dual_tessellation);
+
+    return mTileRearrangement;
+}
```
hryniuk: You could achieve a lot with proper functions' naming instead of putting comments before each call. For example:
```cpp
validateColours(); // areUnique(regionColours);
decomposeBaseTile(base_tile); // convertToPartitionMasksStack(base_tile);
makeDual(); // translateToDualTileParition(baseParition); (and return dualPartition)
```

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
...
+        auto uniform_colour_cube = ImageColour(arma::size(base_tile_cube));
+        uniform_colour_cube.slice(0).fill(regionColours[c].r);
+        uniform_colour_cube.slice(1).fill(regionColours[c].g);
+        uniform_colour_cube.slice(2).fill(regionColours[c].b);
+        uniform_colour_cube.slice(3).fill(regionColours[c].a);
+        ImageStackGrey channel_matches = arma::conv_to<ImageStackGrey>::from(
+            (base_tile_cube == uniform_colour_cube).eval());
+        // Armadillo doesn't have a product or general reducer function,
+        // so we have to use sum instead.
+        const auto not_function = [](bool x) { return !x; };
```
hryniuk: There's `std::logical_not` :).

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
...
+{
+    if (!isNonzero(mTileRearrangement.base))
+    {
+        logError() << "Base tile empty";
+        throw std::runtime_error("Base tile empty");
+    }
+    if (!isBinary(mTileRearrangement.base))
+    {
+        logError() << "Base tile has overlapping regions";
+        throw std::runtime_error("Base tile not correctly partitioned");
```
hryniuk: Error messages should be the same I suppose.

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
...
+    }
+    mTileRearrangement.dual = arma::sum(mTileRearrangement.dualPartition, 2).eval();
+}
+
+std::vector<IVec> TileRearranger::makeTessellationOffsets() const
+{
+    std::vector<IVec> offsets;
+    for (int y = 0; y < 3; y++)
+    {
+        for (int x = 0; x < 3; x++)
```
hryniuk: Why `3`s? It must be named.

Wangscape/image/TileRearranger.cpp
```diff
@@ -0,0 +1,263 @@
...
+        {
+            offsets.push_back((x - 1)*mTileRearrangement.offsetA + (y - 1)*mTileRearrangement.offsetB);
```
hryniuk: Why `-1`?

Wangscape/image/TileRearranger.h
```diff
@@ -0,0 +1,33 @@
...
+public:
+    TileRearranger();
+    ~TileRearranger() = default;
+
+    const TileRearrangement rearrangeTile(const sf::Image& base_tile,
+                                          IVec offset_a,
+                                          IVec offset_b);
+
+    constexpr static unsigned int Corners = 4;
+    std::array<sf::Color, Corners> regionColours;
```
hryniuk: Could it be `private`?

Wangscape/Vector.h
```diff
@@ -0,0 +1,31 @@
...
+
+inline UVec makeUVec(sf::Vector2u v)
+{
+    return UVec(v.x, v.y);
+}
+
+inline DVec makeDVec(sf::Vector2<double> v)
+{
+    return DVec(v.x, v.y);
+}
```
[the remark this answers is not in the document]
serin-delaunay: There might be more `sf::Vector2`/`Vector` conversions when this class is used in the rest of the codebase.

WangscapeTest/TestVector.cpp
```diff
@@ -0,0 +1,60 @@
...
+    EXPECT_EQ(UVec(10, 15), uv * 5);
+
+    EXPECT_EQ(DVec(2.5, 2.75), dv2 / 4);
+}
+
+TEST_F(TestVector, TestVectorConversions)
+{
+    EXPECT_EQ(IVec(2, 3), IVec(uv));
+    EXPECT_EQ(DVec(6, 7), DVec(iv2));
+}
```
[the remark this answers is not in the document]
serin-delaunay: This isn't meant to test whether `vmmlib` works correctly - they have thorough tests. This just checks that expected features (assignment, conversion, vector-vector ops, vector-scalar ops) exist. I tried other libraries as the backend of `Vector` (in particular Armadillo's fixed vector class), and found these basic features were missing or inconvenient to use. If the `Vector` backend changes again, I want these tests in place to make sure the change is transparent.

Wangscape/image/Image.cpp
```diff
@@ -1,13 +1,46 @@
...
+    for (unsigned int s = 0; s < n_steps; s++)
+    {
+        copyRegionBounded(image, displacements.slice(s + 1),
+                          IVec(0, 0), steps[s],
+                          IVec(makeUVec(arma::size(image))));
+    }
+    if (dilate)
+        return arma::max(displacements, 2).eval();
+    else
+        return arma::min(displacements, 2).eval();
```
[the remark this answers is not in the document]
serin-delaunay: `dilatedEroded` is the common implementation of two functions: `eroded` and `dilated`, which both just delegate to this function.

Wangscape/image/Image.cpp
```diff
@@ -30,3 +63,254 @@ ImageGrey imageToGreyscale(const ImageColour & image)
...
+    IVec end_excess;
+    const auto source_end = source_origin + size;
+    const auto target_end = target_origin + size;
+    end_excess.x() = std::max({source_end.x() - (int)source_shape.x(),
+                             target_end.x() - (int)target_shape.x(),
+                             0});
+    end_excess.y() = std::max({source_end.y() - (int)source_shape.y(),
+                             target_end.y() - (int)target_shape.y(),
+                             0});
+    size -= end_excess;
```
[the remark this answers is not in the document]
serin-delaunay: I don't think there's much reason to use these two parts of the function separately. However, everything except the `CopyRegion` call should be moved into a separate function to make testing easier - at the moment this can only be tested by checking the output array.

Wangscape/image/Image.cpp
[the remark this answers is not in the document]
hryniuk: Yes, I've seen it some time after I've put this comment, but, still, I'd change it a bit. Now it's
`dilatedEroded(const ImageGrey &, bool, arma::u8, bool dilate)` and with the last argument equal `false` that's like `dilatedErodedNotDilated` for me and that's why I'd like to have two functions.

With `generateDisplacements()` I've suggested above one of them would be:

```cpp
const auto& steps = use_diagonals ? ORTHODIAGONAL_STEPS : ORTHOGONAL_STEPS;
const auto displacements = generateDisplacements(...);
return arma::max(displacements, 2).eval();
```

and the second one almost the same, but with `arma::min` at the end. Does it sound more reasonable :)?

Wangscape/image/Image.cpp
```diff
@@ -30,3 +63,254 @@ ImageGrey imageToGreyscale(const ImageColour & image)
...
+    border.insert(IVec(start(1), start(0)));
+    const auto& steps = use_diagonals ? ORTHODIAGONAL_STEPS : ORTHOGONAL_STEPS;
+    const auto shape = makeUVec(arma::size(image));
+    const auto in_bounds = [&shape](IVec point)
+    {
+        if (point.x() < 0 || point.y() < 0)
+            return false;
+        if ((unsigned int)point.x() >= shape.x() || (unsigned int)point.y() >= shape.y())
+            return false;
+        return true;
```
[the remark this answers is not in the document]
serin-delaunay: It should probably be a separate function - it's used again in `distances`, and it's a common enough operation. It's also in `sf::Rect<T>::contains()`.

Wangscape/TerrainImages.h
```diff
@@ -1,16 +1,19 @@
...
     std::map<TerrainID, sf::Texture> terrain_textures;
-    TerrainImages() = default;
+    TerrainImages(sf::Vector2u size);
     virtual ~TerrainImages() = default;
-    static void rearrangeTile(sf::Texture& t);
+    void rearrangeTile(sf::Image& base_tile);
     void addTerrain(TerrainID tid, std::string image_filename, std::string json_filename,
                     size_t offset_x, size_t offset_y, sf::Vector2u resolution);
 private:
     std::map<std::string, sf::Image> source_images;
```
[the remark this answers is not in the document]
serin-delaunay: Needs correct capitalisation, too.

Wangscape/image/Image.cpp
serin-delaunay: Can you clarify this?

Wangscape/TerrainImages.cpp
```diff
@@ -46,9 +30,12 @@ void TerrainImages::addTerrain(TerrainID tid, std::string image_filename, std::s
         it = source_images.insert({ filename, img }).first;
     }
     sf::Vector2i offset(offset_x, offset_y);
```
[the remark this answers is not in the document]
serin-delaunay: Yes, making replacing `offset_*` and `resolution` with `Vector` arguments will be part of #46.

Wangscape/image/TileRearranger.cpp
serin-delaunay: We have a lot of blocks like this in the code now - one statement writes to `logError()`, the next throws a `std::runtime_error` with a similar or abbreviated message. It might be useful to have a function which can do both of these things, something like:
```cpp
bool x = false;
if(x)
{
    logAndThrowError<std::runtime_error>() << "Logic failure: " << x;
    // creates temporary variable in this block, throws error with the same message when the variable is destructed.
}
```

Wangscape/image/TileRearranger.h
serin-delaunay: Eventually `TileRearranger` will be run on user-provided partition images, and enforcing the RGBY colour scheme may harm accessibility. For that reason I think it's important for `regionColours` to be mutable. They could be provided in the constructor or `rearrangeTile()`, but I think it's more appropriate for them to be assigned separately. Since it's a `std::array` and `sf::Color` is just a struct, the only harm that could be done is specifying incorrect colours (non-unique or missing). We check for both of those issues before processing, so they don't really need to be hidden behind `get`/`set` methods.

Wangscape/image/TileRearranger.cpp
serin-delaunay: This should really be `for (int y=-1; y<=1; y++)`, with `y` coefficients inside the loop. Same for `x`.

Wangscape/image/Image.cpp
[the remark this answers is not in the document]
serin-delaunay: Yes, that's a good idea. `tessellated` should also delegate to `generateDisplacements()`.

Wangscape/Vector.h
```diff
@@ -0,0 +1,31 @@
...
+
+inline UVec makeUVec(arma::SizeMat v)
+{
+    return UVec(v.n_cols, v.n_rows);
```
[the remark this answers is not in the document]
serin-delaunay: How? `arma::SizeMat` can't be converted to `sf::Vector`.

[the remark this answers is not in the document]
hryniuk: Right, mistake.

Wangscape/image/Image.cpp
hryniuk: I mean defining a function to traverse through matrix and map elements with a given function and neighborhood. I'm not sure if:
* it's worth an effor
* is there an easy way to do so.

So I'm only suggesting it.

I've tried to do it in one of my projects and it works this way:
* I've defined a function, which traverses a given matrix `A`, executing a given function `f` on its elements:

https://github.com/hryniuk/image-processing/blob/19f2fa10441a586657e2f5f44066bb8052116064/op/IndexForEach.hpp#L14-L26

* I've defined an operators that can be passed to the above function, which are calculating a value based on neighborhood (`surr`oundings):

https://github.com/hryniuk/image-processing/blob/19f2fa10441a586657e2f5f44066bb8052116064/op/Mean.hpp#L16-L31

And this inner `for` loop can be generalized as a function that executes a value with a given neighborhood.

**But**:
* for more complex operations it could be harder to read and analyze
* for simple things there is `arma::conv2`

sooo... ¯\\\_(ツ)_/¯.

Wangscape/image/TileRearranger.cpp
hryniuk: For me better would be treating these functions as assertions and throwing a custom exceptions. The whole part would be just:

```cpp
try
{
    validateRearrangment(mTileRearrangement);
}
catch (const InvalidRearrangementTile& e)
{
    logError() << e.what()
}
```

where `validateRearrangment` calls `assertConnected`, `assertNonZero` and so on. These functions should throw assertion-specific exceptions, like `TileNotConnected`, `EmptyTileRegion`, `IncorrectTilePartition`.

Wangscape/image/TileRearranger.h
hryniuk: > Since it's a std::array and sf::Color is just a struct, the only harm that could be done is specifying incorrect colours (non-unique or missing).

I sounds like a separate entity - `class` that checks all these preconditions in its constructor.

Wangscape/image/Image.cpp
serin-delaunay: I think it will worthwhile to split `distance()` up and try to unify its implementation with `isConnected()`. Some of the ideas in #37 also require similar operations. However, this is fundamentally a graph algorithm (breadth first search) - the order of processing is determined by the numerical values of the input arrays, not their size. Trying to iterate over the whole matrix row-by-row would obscure what the algorithm does, rather than clarify it.

serin-delaunay:
From Gitter:
https://gitter.im/Wangscape/Lobby?at=59b6d11d319100804e212ebf
> The padding ensures that none of the pixels in the base mask are adjacent to the edge of the image.
That one. By "image" you mean base tile mask in the center?
Also, there are parts of mask (white pixels, 1's) in corner images (3 and 7).
```
1 2 3
4 5 6
7 8 9
```
Is it intentional?

https://gitter.im/Wangscape/Lobby?at=59b6d360bc464729741ee399
> I could pad the base tile so that there are more pixels visible in 1,3,7,9 - this would be automatically cut off in the dual tile

https://gitter.im/Wangscape/Lobby?at=59b6d85b319100804e215374
> Apart from padding the base tile from the start, I could add a separate line showing the extra padding before the tessellation diagram

> self-dual tile is one that has the same border as its dual tile?

https://gitter.im/Wangscape/Lobby?at=59b6deeb614889d475e5f9af
> Ok, finished. Code follows that description, so it's not so hard to get as I thought :). Four more comments on "Rearranging source tiles" and I'll try to add another PR:
> * "chess board-style display"; it means just a grid?
> * `void.png` and `void-corners.png` seems to be black on my (bright) screen, contrast increase would help
> * masks overlapping - I didn't get it at first; if I'm right, it means that sum of two masks should be binary, yes? Could you emphasize it? Also, there's "with each translated copy in a different mask" statement. As far as I understand, they're just base tile mask translated by vectors provided by the user. Am I right?
> * some of keywords should be **emphasized**, e.g. "regions", "base tile" or "dual tile mask". It will be easier to read.

Wangscape/TerrainImages.cpp
```diff
@@ -46,9 +30,12 @@ void TerrainImages::addTerrain(TerrainID tid, std::string image_filename, std::s
...
+    sf::Image tile;
+    tile.create(resolution.x, resolution.y);
+    tile.copy(it->second, 0,0, sf::IntRect(offset, sf::Vector2i(resolution)));
+    rearrangeTile(tile);
     sf::Texture tex;
-    if (!tex.loadFromImage((*it).second, sf::IntRect(offset, sf::Vector2i(resolution))))
+    if (!tex.loadFromImage(tile))
         throw std::runtime_error("Couldn't make texture from image");
```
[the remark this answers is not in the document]
serin-delaunay: I've extracted the BFS component of `distances` and `isConnected` into a function, but it uses four different callbacks. Perhaps it would be better to make it a functor with 4 protected pure virtual functions.

Wangscape/image/Image.cpp
serin-delaunay: I've extracted the BFS component of `distances` and `isConnected` into a function, but it uses four different callbacks. Perhaps it would be better to make it a functor with 4 protected pure virtual functions.

PR #196, opened 2017-09-09
Reseed border and combiner module groups on load

### LINKED ISSUE #193: Call setSeeds() on border and combiner module groups during initialisation
Currently border and combiner module groups are not being reseeded on startup, which means that the seeds used are user-specified or zero. If the same module group is used for horizontal and vertical borders, they will not differ. They will also not change when the program is re-run. This can be seen in the current `example2` (check `g.s.border_v.OUTPUT.TL.png` and `g.s.border_h.OUTPUT.TL.png`).

To fix this, add `setSeeds()` calls when loading the combiner and border module groups.

serin-delaunay:
Fixes #193.

Required for https://github.com/Wangscape/Wangscape-examples to produce correct output.

### DIFF
--- a/Wangscape/noise/ModuleGroupContainer.h
+++ b/Wangscape/noise/ModuleGroupContainer.h
@@ -3,6 +3,7 @@
 #include <map>
 #include <memory>
 #include <string>
+#include <random>
 
 #include <boost/optional.hpp>
 
@@ -17,8 +18,8 @@ class ModuleGroupContainer
 public:
     explicit ModuleGroupContainer(std::string module_group_description_);
     virtual ~ModuleGroupContainer() = default;
-    void addSpecificModuleGroup(T key, std::string filename);
-    void tryAddDefaultModuleGroup(T key, boost::optional<std::string> filename);
+    void addSpecificModuleGroup(T key, std::string filename, std::mt19937& rng);
+    void tryAddDefaultModuleGroup(T key, boost::optional<std::string> filename, std::mt19937& rng);
     ModuleGroup& at(const T& key);
 private:
     typedef std::map<T, std::shared_ptr<ModuleGroup>> ModuleGroups;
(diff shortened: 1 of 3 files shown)


Wangscape/noise/ModuleGroupContainer.h
```diff
@@ -17,8 +18,8 @@ class ModuleGroupContainer
 public:
     explicit ModuleGroupContainer(std::string module_group_description_);
     virtual ~ModuleGroupContainer() = default;
-    void addSpecificModuleGroup(T key, std::string filename);
-    void tryAddDefaultModuleGroup(T key, boost::optional<std::string> filename);
+    void addSpecificModuleGroup(T key, std::string filename, std::mt19937& rng);
+    void tryAddDefaultModuleGroup(T key, boost::optional<std::string> filename, std::mt19937& rng);
```
hryniuk: I don't like this interface. I see two problems here:
* as this is `ModuleGroupContainer` and it has `*add*ModuleGroup` methods, I'd expect that they accept already loaded `ModuleGroup`, not filename to it (but it's a different issue)
* as you're calling `setSeeds` after every `insert` and the issue you want to resolve is called "Call setSeeds() on border and combiner module groups **during initialisation**", the best solution in my opinion would be calling `setSeeds` in the `ModuleGroup`'s constructor. By doing so, you've got guarantee that it is always called. Good idea could be also passing seed to it, as that class doesn't need RNG after generating seed.

Do you agree?

serin-delaunay: * I could rename them to `*load*ModuleGroup` or add an intermediate function, but there's no reason to keep a `ModuleGroup` object around after it's added to a `ModuleGroupContainer`. Duplicate module groups must be fully independent copies (so that they can be used together with different seeds), and libnoise `Module`s aren't copyable. This means a `ModuleGroup` must be loaded every time it's added to a `ModuleGroupContainer`.
* The `setSeeds()` call could be moved to `loadModuleGroup()`. I don't think it should be moved as far as the `ModuleGroup` constructor or the `EncodedModuleGroup::decode()` call.

hryniuk: Looks good :).

PR #205, opened 2017-09-16
Allow comments in module JSON

V1shvesh:


serin-delaunay:
That's great, but it only works for the `CornerCombiner` module. Could you also enable comments for the rest of the modules? If you put the `"comment"` schema inside `#/definitions`, then this would be as simple as adding `"comment": { "$ref": "#/definitions/comment"}` to each of the module schemas.

V1shvesh:
Sure. Thanks for the help!

V1shvesh:
Comment for `"cache"' required??

serin-delaunay:
Yes.

V1shvesh:
Check it now perhaps...

serin-delaunay:
Thanks! I see why you asked about those four modules - they haven't been added to the `#/modules` schema. If you fix that bug, this will be ready to merge.

V1shvesh:
Done! :+1:

doc/schemas/module_group_schema.json
```diff
@@ -52,6 +55,7 @@
           {"$ref": "#/definitions/pow"},
           {"$ref": "#/definitions/quadrant_selector"},
           {"$ref": "#/definitions/ridged_multi"},
+          {"$ref": "#/definitions/reciprocal"},
```
serin-delaunay: `reciprocal` should come before `ridged_multi`.

V1shvesh: Didn't know how I missed that...

V1shvesh:
Reordered `reciprocal`

serin-delaunay:
Thanks, great work. Merged.

V1shvesh:
Thanks to you as well! This was absolutely amazing!

PR #206, opened 2017-09-28
Add libarmadillo-dev to pre-requisite installation instructions

bluehood:
Just found out the hard way that `libarmadillo-dev` is missing from the pre-requisite installation instructions for Ubuntu linux.

(by the way, how do you manage to have my shell print that annoying warning message about CRLF and LF everytime I execute a command inside the Wangscape repo? It's incredibly weird and amazing at the same time)

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ First, install necessary packages with `apt`:
 ```shell
 apt install cmake g++ libsfml-dev libboost-system-dev \
     libboost-filesystem-dev libboost-program-options-dev \
-    qt5-default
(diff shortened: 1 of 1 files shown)


serin-delaunay:
Thanks for the PR!

I thought it was just me seeing the CRLF warnings, and they were only appearing in gitextensions logs. If they're appearing in your console, and frequently, it ought to be fixed. Could you post a copy of the warning text?

bluehood:
@serin-delaunay the CRLF warnings are now issue #208

